PR #3, opened 2020-11-24
Add benchmarks for opening/closing shared/confined native/heap memory segments

chrisvest:
The numbers currently say that shared segments are quite expensive to open and close, regardless of how many threads running in the JVM. JFR profiling indicates that the `HandshakeAllThreads` VM operation is taking up most of the time, but JFR is not able to pin-point it further than that, since VM operations don't carry stack traces.

```
Benchmark                                   (size)  (unrelatedThreads)  Mode  Cnt      Score     Error  Units
MemorySegmentCloseBenchmark.heapConfined        16                   0  avgt   50     33,097 ±   1,914  ns/op
MemorySegmentCloseBenchmark.heapConfined        16                  10  avgt   50     33,471 ±   1,904  ns/op
MemorySegmentCloseBenchmark.heapConfined        16                 100  avgt   50     33,289 ±   1,962  ns/op
MemorySegmentCloseBenchmark.heapShared          16                   0  avgt   50  19675,203 ± 343,392  ns/op
MemorySegmentCloseBenchmark.heapShared          16                  10  avgt   50  19396,607 ± 337,509  ns/op
MemorySegmentCloseBenchmark.heapShared          16                 100  avgt   50  19366,129 ± 308,155  ns/op
MemorySegmentCloseBenchmark.nativeConfined      16                   0  avgt   50    256,534 ±   3,996  ns/op
MemorySegmentCloseBenchmark.nativeConfined      16                  10  avgt   50    256,942 ±   3,888  ns/op
MemorySegmentCloseBenchmark.nativeConfined      16                 100  avgt   50    255,618 ±   4,298  ns/op
MemorySegmentCloseBenchmark.nativeShared        16                   0  avgt   50  19328,519 ± 105,996  ns/op
MemorySegmentCloseBenchmark.nativeShared        16                  10  avgt   50  18997,370 ± 249,796  ns/op
MemorySegmentCloseBenchmark.nativeShared        16                 100  avgt   50  18775,365 ± 234,154  ns/op
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -386,6 +386,7 @@
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
       <version>3.18.0</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
--- a/src/test/java/io/netty/buffer/api/BufTest.java
+++ b/src/test/java/io/netty/buffer/api/BufTest.java
@@ -1513,7 +1513,7 @@ public class BufTest {
     }
 
     @ParameterizedTest
-    @MethodSource("poolingAllocators")
+    @MethodSource("allocators")
     public void pooledBuffersMustResetStateBeforeReuse(Fixture fixture) {
         try (Allocator allocator = fixture.createAllocator();
              Buf expected = allocator.allocate(8)) {
--- a/src/test/java/io/netty/buffer/api/benchmarks/MemSegBufAccessBenchmark.java
+++ b/src/test/java/io/netty/buffer/api/benchmarks/MemSegBufAccessBenchmark.java
@@ -13,8 +13,10 @@
 * License for the specific language governing permissions and limitations
 * under the License.
 */
-package io.netty.buffer.api;
+package io.netty.buffer.api.benchmarks;
 
+import io.netty.buffer.api.Allocator;
+import io.netty.buffer.api.Buf;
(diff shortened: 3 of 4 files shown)


chrisvest:
The thread-local handshakes dominate:
[flame-cpu-forward.svg.gz](https://github.com/netty/netty-incubator-buffer-api/files/5590242/flame-cpu-forward.svg.gz)
Screenshot from 2020-11-24 14-52-24

chrisvest:
Also notifying @mcimadamore who might find it interesting.

franz1981:
That's the sad reality although I believe there must (fingers crossed) be some param to tune the spinning to save pthread cond to be issued. Need to play a bit on this by my own before suggesting anything useful :P

chrisvest:
Our plan so far is to only use shared segments when necessary, e.g. with pooling allocators.

chrisvest:
What I think is not captured by this benchmark, is that the handshake can cause other threads to have their currently running Java code to be de-optimised. So closing a shared segment can be slow not just for the calling thread, but also for other threads in the system.
See https://github.com/openjdk/panama-foreign/blob/65b410dcfcf68e4c7e21195bc4a479703a48422b/src/hotspot/share/prims/scopedMemoryAccess.cpp#L100-L112

mcimadamore:
Thanks for bringing this to my attention. We'll take a look.

Issue #6, opened 2020-11-26
Test failure in build: BufTest.relativeReadOfUnsignedByteMustNotBoundsCheckWhenReadOffsetAndSizeIsEqualToWriteOffset


chrisvest:
https://github.com/netty/netty-incubator-buffer-api/actions/runs/385257419

```
2020-11-26T13:07:26.5820359Z [INFO] -------------------------------------------------------
2020-11-26T13:07:26.5821826Z [INFO]  T E S T S
2020-11-26T13:07:26.5838079Z [INFO] -------------------------------------------------------
2020-11-26T13:07:27.3085876Z [ERROR] WARNING: Using incubator modules: jdk.incubator.foreign
2020-11-26T13:07:31.7609030Z [INFO] Running io.netty.buffer.api.BufTest
2020-11-26T13:08:03.2065216Z [INFO] Running io.netty.buffer.api.BufTest$CleanerTests
2020-11-26T13:08:04.8472333Z [ERROR] Tests run: 33621, Failures: 0, Errors: 1, Skipped: 2, Time elapsed: 0.039 s <<< FAILURE! - in io.netty.buffer.api.BufTest$CleanerTests
2020-11-26T13:08:04.8533154Z [ERROR] io.netty.buffer.api.BufTest.relativeReadOfUnsignedByteMustNotBoundsCheckWhenReadOffsetAndSizeIsEqualToWriteOffset(Fixture)[124]  Time elapsed: 0.013 s  <<< ERROR!
2020-11-26T13:08:04.8538438Z java.lang.IllegalStateException: This segment is already closed
2020-11-26T13:08:04.8541560Z 	at jdk.incubator.foreign/jdk.internal.foreign.AbstractMemorySegmentImpl.checkValidState(AbstractMemorySegmentImpl.java:414)
2020-11-26T13:08:04.8546105Z 	at jdk.incubator.foreign/jdk.internal.foreign.AbstractMemorySegmentImpl.close(AbstractMemorySegmentImpl.java:335)
2020-11-26T13:08:04.8548951Z 	at io.netty.buffer.api.memseg.MemSegBuf.lambda$static$0(MemSegBuf.java:47)
2020-11-26T13:08:04.8551444Z 	at io.netty.buffer.api.SizeClassedMemoryPool.dispose(SizeClassedMemoryPool.java:127)
[...]

PR #7, opened 2020-11-30
@chrisvest Capture build artifacts for failed builds

chrisvest:


chrisvest:
The build failures reported in #6 have stopped occurring, so I'll just merge this, and then hopefully we'll be more ready if they start showing up again.

PR #10, opened 2020-12-02
Add benchmarks for ByteIterator

chrisvest:
Motivation:
Capture the performance characteristics of this primitive for various buffer implementations.

Modification:
Add a benchmark that iterate 4KiB buffers forwards, and backwards, on various buffer implementations.

Result:
Another aspect of the implementation covered by benchmarks.
Turns out the composite iterators a somewhat slow.

### DIFF
--- a/src/test/java/io/netty/buffer/api/benchmarks/ByteIterationBenchmark.java
+++ b/src/test/java/io/netty/buffer/api/benchmarks/ByteIterationBenchmark.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2019 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


chrisvest:
```
Benchmark                                    (type)  Mode  Cnt     Score    Error  Units
ByteIterationBenchmark.sum                     heap  avgt   50   682,136 ±  3,934  ns/op
ByteIterationBenchmark.sum                   direct  avgt   50   621,579 ±  3,009  ns/op
ByteIterationBenchmark.sum           composite-heap  avgt   50  3469,898 ± 43,624  ns/op
ByteIterationBenchmark.sum         composite-direct  avgt   50  2703,371 ± 53,141  ns/op
ByteIterationBenchmark.sumReverse              heap  avgt   50   438,394 ±  7,681  ns/op
ByteIterationBenchmark.sumReverse            direct  avgt   50   463,282 ±  7,074  ns/op
ByteIterationBenchmark.sumReverse    composite-heap  avgt   50  2713,186 ± 48,320  ns/op
ByteIterationBenchmark.sumReverse  composite-direct  avgt   50  2735,066 ± 91,636  ns/op
```

Interesting how the reverse iteration is faster than forward iteration.

PR #11, opened 2020-12-02
Turn ByteIterator into ByteCursor

### LINKED ISSUE #8: Change ByteIterator to a ByteCursor API
This will remove half of the "has next" checks and likely make the use sites easier for the compiler to optimise.

chrisvest:
Motivation:
Cursors are better than iterators in that they only need to check boundary conditions once per iteration, when processed in a loop.
This should make them easier for the compiler to optimise.

Modification:
Change the ByteIterator to a ByteCursor. The API is almost the same, but with a few subtle differences in semantics.
The primary difference is that the cursor movement and boundary condition checking and position movement happen at the same time, and do not need to occur when the values are fetched out of the cursor.
An iterator, on the other hand, needs to throw an exception if "next" is called too many times.

Result:
Simpler code, and faster code as well.

This PR:

```
Benchmark                                    (type)  Mode  Cnt     Score    Error  Units
ByteIterationBenchmark.sum                     heap  avgt   50   478,618 ±  4,147  ns/op
ByteIterationBenchmark.sum                   direct  avgt   50   476,447 ±  6,019  ns/op
ByteIterationBenchmark.sum           composite-heap  avgt   50  2986,791 ± 73,203  ns/op
ByteIterationBenchmark.sum         composite-direct  avgt   50  2670,200 ± 48,309  ns/op
ByteIterationBenchmark.sumReverse              heap  avgt   50   390,934 ±  5,148  ns/op
ByteIterationBenchmark.sumReverse            direct  avgt   50   392,272 ±  6,538  ns/op
ByteIterationBenchmark.sumReverse    composite-heap  avgt   50  2770,515 ± 67,099  ns/op
ByteIterationBenchmark.sumReverse  composite-direct  avgt   50  2680,819 ± 61,209  ns/op
```

And previously, with the iterator pattern:

```
Benchmark                                    (type)  Mode  Cnt     Score    Error  Units
ByteIterationBenchmark.sum                     heap  avgt   50   682,136 ±  3,934  ns/op
ByteIterationBenchmark.sum                   direct  avgt   50   621,579 ±  3,009  ns/op
ByteIterationBenchmark.sum           composite-heap  avgt   50  3469,898 ± 43,624  ns/op
ByteIterationBenchmark.sum         composite-direct  avgt   50  2703,371 ± 53,141  ns/op
ByteIterationBenchmark.sumReverse              heap  avgt   50   438,394 ±  7,681  ns/op
ByteIterationBenchmark.sumReverse            direct  avgt   50   463,282 ±  7,074  ns/op
ByteIterationBenchmark.sumReverse    composite-heap  avgt   50  2713,186 ± 48,320  ns/op
ByteIterationBenchmark.sumReverse  composite-direct  avgt   50  2735,066 ± 91,636  ns/op
```

So reverse iteration is still faster, somehow. Composite got a bit slower, but it ought to be possible to improve on that. There's no good reason, I think, that composite should have a 10x overhead on this.
For non-composite, things are faster overall.


chrisvest:
@franz1981 what do you think of this?

src/main/java/io/netty/buffer/api/CompositeBuf.java
```diff
@@ -317,61 +316,63 @@ public ByteIterator iterate(int fromOffset, int length) {
...
-                    throw new NoSuchElementException();
+            public boolean nextLong() {
+                if (bytesLeft() >= Long.BYTES) {
+                    if (cursor.nextLong()) {
+                        longValue = cursor.getLong();
+                        index += Long.BYTES;
+                    } else {
+                        longValue = nextLongFromBytes(); // Leave index increments to 'nextByte'
+                    }
+                    return true;
```
chrisvest, line 345: Maybe if `ByteCursor` have `setLong` and `setByte` that overwrite the current values, then we could make it the common case to just directly delegate the `nextLong` call, and avoid caching the values in the composite byte cursor. If we can remove `bytesLeft` from the hot path, then we can make it computed and avoid maintaining the `index` field as well.

franz1981:
First look is good, I am a big fan of cursors API vs iteration..tomorrow I will take a better look

franz1981:
GC apart would be nice to check how the swar indexOf compare vs something using cursors.

src/main/java/io/netty/buffer/api/Buf.java
```diff
@@ -320,11 +320,11 @@ default ByteIterator iterate() {
      * @param fromOffset The offset into the buffer where iteration should start.
      *                  The first byte read from the iterator will be the byte at this offset.
      * @param length The number of bytes to iterate.
-     * @return A {@link ByteIterator} for the given stretch of bytes of this buffer.
+     * @return A {@link ByteCursor} for the given stretch of bytes of this buffer.
      * @throws IllegalArgumentException if the length is negative, or if the region given by the {@code fromOffset} and
      * the {@code length} reaches outside of the bounds of this buffer.
      */
-    ByteIterator iterate(int fromOffset, int length);
+    ByteCursor iterate(int fromOffset, int length);
```
normanmaurer: should we change the method name to `cursor` ?

normanmaurer: Same comment for the other methods.

src/main/java/io/netty/buffer/api/ByteCursor.java
```diff
@@ -17,48 +17,55 @@
...
-     * bytes that might form a tail in the iterator.
+     * bytes that might form a tail in the cursor.
+     * <p>
+     * Also note that this method will not influence what is returned the {@link #getByte()} method.
      *
-     * @return {@code true} if a call to {@link #nextLong()} would succeed, otherwise {@code false}.
+     * @return {@code true} if the cursor read 8 bytes and moved forward, otherwise {@code false}.
      */
-    boolean hasNextLong();
+    boolean nextLong();
```
normanmaurer: I wonder if we should rename this to make it more clear that exactly happens here (Like moving the cursor forward).

src/main/java/io/netty/buffer/api/ByteCursor.java
```diff
@@ -17,48 +17,55 @@
...
      *
-     * @return {@code true} if the next call to {@link #nextByte()} would succeed, otherwise {@code false}.
+     * @return {@code true} if the cursor read a byte and moved forward, otherwise {@code false}.
      */
-    boolean hasNextByte();
+    boolean nextByte();
 
     /**
-     * Read and return the next byte, and move the iterator position] forward by one byte.
+     * Return the last byte that was read by {@link #nextByte()}.
```
normanmaurer: What is the contract if you call `getByte()` without `nextByte()` ?

src/main/java/io/netty/buffer/api/CompositeBuf.java
normanmaurer: I would prefer to not introduce `set*` methods

normanmaurer (changes requested):

src/main/java/io/netty/buffer/api/Buf.java
chrisvest: Considered it, but `cursor` is not a verb. I've used `open` or `openCursor` previously. How does that sound?

src/main/java/io/netty/buffer/api/ByteCursor.java
chrisvest: Hmm… `readLong`?

src/main/java/io/netty/buffer/api/ByteCursor.java
chrisvest: Specified this in the javadocs.

src/main/java/io/netty/buffer/api/CompositeBuf.java
chrisvest: Do you object to the functionality of overriding the cursor state, or the method names, or both?

normanmaurer: override the state... I would prefer to keep the state the cursors own business

src/main/java/io/netty/buffer/api/Buf.java
normanmaurer: `openCursor` sounds good imho

src/main/java/io/netty/buffer/api/ByteCursor.java
normanmaurer: yeah maybe... its just a bit odd in the sense that `readLong` method on the `ByteBuf` actually returns a long ..

src/main/java/io/netty/buffer/api/ByteCursor.java
normanmaurer: Can you also specify what will happen if there is a mismatch between the number of calls... like what happens if you call `nextByte()` 2 times but `getByte()` only once (after the second call). Also what happens if you call `nextByte()` once but `getByte()` twice.

chrisvest:
@normanmaurer I tried to squeeze a bit more performance out of the composite ByteCursor. It helped, but not nearly as much as I had hoped.

Before:

```
Benchmark                                    (type)  Mode  Cnt     Score    Error  Units
ByteIterationBenchmark.sum                     heap  avgt   10   490,014 ±  6,874  ns/op
ByteIterationBenchmark.sum                   direct  avgt   10   489,624 ±  3,916  ns/op
ByteIterationBenchmark.sum           composite-heap  avgt   10  3707,880 ± 82,760  ns/op
ByteIterationBenchmark.sum         composite-direct  avgt   10  3731,500 ± 66,111  ns/op
ByteIterationBenchmark.sumReverse              heap  avgt   10   401,745 ±  3,869  ns/op
ByteIterationBenchmark.sumReverse            direct  avgt   10   398,329 ±  2,268  ns/op
ByteIterationBenchmark.sumReverse    composite-heap  avgt   10  3510,920 ± 80,826  ns/op
ByteIterationBenchmark.sumReverse  composite-direct  avgt   10  3229,770 ± 72,151  ns/op
```

After:

```
Benchmark                                    (type)  Mode  Cnt     Score    Error  Units
ByteIterationBenchmark.sum                     heap  avgt   50   478,618 ±  4,147  ns/op
ByteIterationBenchmark.sum                   direct  avgt   50   476,447 ±  6,019  ns/op
ByteIterationBenchmark.sum           composite-heap  avgt   50  2986,791 ± 73,203  ns/op
ByteIterationBenchmark.sum         composite-direct  avgt   50  2670,200 ± 48,309  ns/op
ByteIterationBenchmark.sumReverse              heap  avgt   50   390,934 ±  5,148  ns/op
ByteIterationBenchmark.sumReverse            direct  avgt   50   392,272 ±  6,538  ns/op
ByteIterationBenchmark.sumReverse    composite-heap  avgt   50  2770,515 ± 67,099  ns/op
ByteIterationBenchmark.sumReverse  composite-direct  avgt   50  2680,819 ± 61,209  ns/op
```

normanmaurer:
@chrisvest but this before / after is still for iterator vs cursor, correct ?

chrisvest:
@normanmaurer No, that one is before/after https://github.com/netty/netty-incubator-buffer-api/pull/11/commits/ae22e48e339cf7b061810a69934abf7a0e7c1a8b

chrisvest:
Merging this because all comments have been addressed.

PR #12, opened 2020-12-03
Make it possible to extend composite buffers after creation

chrisvest:
Motivation:
Composite buffers are uniquely positioned to be able to extend their underlying storage relatively cheaply.
This fact is relied upon in a couple of buffer use cases within Netty, that we wish to support.

Modification:
Add a static `extend` method to Allocator, so that the CompositeBuf class can remain internal.
The `extend` method inserts the extension buffer at the end of the composite buffer as if it had been included from the start.
This involves checking offsets and byte order invariants.
We also require that the composite buffer be in an owned state.

Result:
It's now possible to extend a composite buffer with a specific buffer, after the composite buffer has been created.

### DIFF
--- a/src/main/java/io/netty/buffer/api/Allocator.java
+++ b/src/main/java/io/netty/buffer/api/Allocator.java
@@ -112,6 +112,27 @@ public interface Allocator extends AutoCloseable {
     }
 
     /**
+     * Extend the given composite buffer with the given extension buffer.
+     * This works as if the extension had originally been included at the end of the list of constituent buffers when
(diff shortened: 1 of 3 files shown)


chrisvest:
@normanmaurer I found that only one of the `CompositeByteBuf.addComponent` overloads were really used in Netty, and the semantics of the `extend()` method here matches that one, AFAICT.

chrisvest:
I'm going to merge this, and then we can talk about moving the API later.

PR #13, opened 2020-12-04
Make Buf.send() faster, again

chrisvest:
Motivation:
This will likely be a somewhat common operation, as buffers move between eventloop and worker threads, so it's important to have an understanding of how it performs.

Modification:
Add a benchmark that specifically targets the send() operation on buffers.
Use a dedicated transfer thread to facilitate the handoff of thread-confined memory segments, from one thread to another.
By using a separate thread for this, we no longer need to create and close a shared segment, sparing us of the thread-local handshake costs related to that.

Result:
We got benchmark numbers that clearly show the cost of confinement transfer, and sending a thread-confined buffer from one thread to another is now about twice as fast.

Before:

```
Benchmark                    Mode  Cnt   Score   Error  Units
SendBenchmark.sendNonPooled  avgt   50  31,348 ± 0,406  us/op
SendBenchmark.sendPooled     avgt   50   8,626 ± 0,165  us/op

Benchmark                                                  (workload)  Mode  Cnt   Score   Error  Units
MemorySegmentClosedByCleanerBenchmark.cleanerClose              heavy  avgt   50  13,121 ± 0,806  us/op
MemorySegmentClosedByCleanerBenchmark.cleanerClose              light  avgt   50   2,269 ± 0,462  us/op
MemorySegmentClosedByCleanerBenchmark.explicitClose             heavy  avgt   50  36,795 ± 0,666  us/op
MemorySegmentClosedByCleanerBenchmark.explicitClose             light  avgt   50   1,291 ± 0,063  us/op
MemorySegmentClosedByCleanerBenchmark.explicitPooledClose       heavy  avgt   50  11,722 ± 0,231  us/op
MemorySegmentClosedByCleanerBenchmark.explicitPooledClose       light  avgt   50   1,063 ± 0,011  us/op
```

After:

```
Benchmark                    Mode  Cnt   Score   Error  Units
SendBenchmark.sendNonPooled  avgt   50  15,788 ± 0,206  us/op
SendBenchmark.sendPooled     avgt   50   8,475 ± 0,072  us/op

Benchmark                                                  (workload)  Mode  Cnt   Score   Error  Units
MemorySegmentClosedByCleanerBenchmark.cleanerClose              heavy  avgt   50  26,931 ± 1,001  us/op
MemorySegmentClosedByCleanerBenchmark.cleanerClose              light  avgt   50   2,430 ± 0,513  us/op
MemorySegmentClosedByCleanerBenchmark.explicitClose             heavy  avgt   50  24,413 ± 0,237  us/op
MemorySegmentClosedByCleanerBenchmark.explicitClose             light  avgt   50   1,142 ± 0,018  us/op
MemorySegmentClosedByCleanerBenchmark.explicitPooledClose       heavy  avgt   50  11,858 ± 0,216  us/op
MemorySegmentClosedByCleanerBenchmark.explicitPooledClose       light  avgt   50   1,071 ± 0,009  us/op
```

### DIFF
--- a/src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
+++ b/src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
@@ -27,6 +27,9 @@ import jdk.incubator.foreign.MemorySegment;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.NoSuchElementException;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.SynchronousQueue;
 
 import static jdk.incubator.foreign.MemoryAccess.getByteAtOffset;
 import static jdk.incubator.foreign.MemoryAccess.getCharAtOffset;
@@ -45,6 +48,22 @@ import static jdk.incubator.foreign.MemoryAccess.setShortAtOffset;
 
 class MemSegBuf extends RcSupport<Buf, MemSegBuf> implements Buf {
     static final Drop<MemSegBuf> SEGMENT_CLOSE = buf -> buf.seg.close();
+
+    private static final BlockingQueue<Runnable> QUEUE = new SynchronousQueue<>();
+    private static final Thread TRANSFER_THREAD = new Thread(() -> {
+        try {
+            for (;;) {
+                QUEUE.take().run();
+            }
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    });
+    static {
+        TRANSFER_THREAD.setDaemon(true);
+        TRANSFER_THREAD.start();
+    }
+
     private final AllocatorControl alloc;
     private final boolean isSendable;
     private MemorySegment seg;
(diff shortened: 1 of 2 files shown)


chrisvest:
The `cleanerClose/heavy` case became slower, I think, because now it doesn't have much advantage compared to explicitly closed. Basically, those two cases aligned, while previously the cleaner case could upgrade the segment to a shared one and then forego its costs of closing.

chrisvest:
I was worried that the transfer thread was perhaps not going to be a scalable solution on a loaded system, but it turns out that the performance benefit is even more pronounced then.

When running the benchmark with `-t 8`:

Before

```
Benchmark                    Mode  Cnt    Score   Error  Units
SendBenchmark.sendNonPooled  avgt   50  419,389 ± 6,732  us/op
SendBenchmark.sendPooled     avgt   50   17,409 ± 0,303  us/op
```

After

```
Benchmark                    Mode  Cnt   Score   Error  Units
SendBenchmark.sendNonPooled  avgt   50  42,229 ± 1,884  us/op
SendBenchmark.sendPooled     avgt   50  18,153 ± 0,347  us/op
```

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -45,6 +48,22 @@
...
     static final Drop<MemSegBuf> SEGMENT_CLOSE = buf -> buf.seg.close();
+
+    private static final BlockingQueue<Runnable> QUEUE = new SynchronousQueue<>();
+    private static final Thread TRANSFER_THREAD = new Thread(() -> {
+        try {
+            for (;;) {
+                QUEUE.take().run();
+            }
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
```
normanmaurer, line 59: should we just call `Thread.currentThread().interrupt()` and return ?

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -45,6 +48,22 @@
...
+        }
+    });
+    static {
+        TRANSFER_THREAD.setDaemon(true);
```
normanmaurer, line 63: I can just tell you from experience that people sometimes are very sensitive to threads that are created in the background and basically never exists. ..

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -45,6 +48,22 @@
...
+        TRANSFER_THREAD.start();
```
normanmaurer, line 64: You should ensure you reset the class loader to not leak at the end 

https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/ObjectCleaner.java#L111

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -707,18 +726,37 @@ public Buf setDouble(int woff, double value) {
...
+                        copy.order = outer.order;
+                        copy.roff = outer.roff;
+                        copy.woff = outer.woff;
+                        return copy;
+                    } catch (Exception e) {
+                        throw new RuntimeException(e);
+                    }
+                }
+            };
+        } else {
```
normanmaurer, line 748: nit: you can remove the `else` as you return before.

src/test/java/io/netty/buffer/api/benchmarks/SendBenchmark.java
```diff
@@ -0,0 +1,68 @@
...
+@Warmup(iterations = 10, time = 1)
+@Measurement(iterations = 10, time = 1)
+@Fork(value = 5, jvmArgsAppend = { "-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints" })
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+public class SendBenchmark {
+    private static final Allocator nonPooled = Allocator.heap();
+    private static final Allocator pooled = Allocator.pooledHeap();
+    private static final Function<Send<Buf>, Send<Buf>> bounce = send -> {
```
normanmaurer: nit: consider using upper-case for the names to be inlines with the JLS recommendation of static fields

src/test/java/io/netty/buffer/api/benchmarks/SendBenchmark.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2019 The Netty Project
```
normanmaurer: 2020

normanmaurer (changes requested):

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
chrisvest: Do you have an alternative in mind?

chrisvest:
@normanmaurer I'll split out the benchmark in a separate PR, since that'll be less controversial.

chrisvest:
Closing this because it becomes a non-issue (no more thread-ownership transitions) when we remove thread-confinement in #21

PR #16, opened 2020-12-10
Add a Buf.bifurcate method

chrisvest:
Motivation:
There are use cases that involve accumulating data into a buffer, then carving out prefix slices and sending them off on their own journey for further processing.

Modification:
Add a Buf.bifurcate API, that split a buffer, and its ownership, in two.
Internally, the API will inject and maintain an atomically reference counted Drop instance, so that the original memory segment is not released until all bifurcated parts are closed.
This works particularly well for composite buffers, where only the buffer (if any) wherein the bifurcation point lands, will actually have its memory split. A composite buffer can otherwise just crack its buffer array in two.

Result:
We now have a safe way of breaking the single ownership of some memory into multiple parts, that can be sent and owned independently.

### DIFF
--- a/src/main/java/io/netty/buffer/api/Buf.java
+++ b/src/main/java/io/netty/buffer/api/Buf.java
@@ -373,4 +373,48 @@ public interface Buf extends Rc<Buf>, BufAccessors {
      * That is, if {@link #countBorrows()} is not {@code 0}.
      */
     void ensureWritable(int size);
+
+    /**
+     * Split the buffer into two, at the {@linkplain #writerOffset() write offset} position.
+     * <p>
+     * The region of this buffer that contain the read and readable bytes, will be captured and returned in a new
+     * buffer, that will hold its own ownership of that region. This allows the returned buffer to be indepentently
+     * {@linkplain #send() sent} to other threads.
+     * <p>
+     * The returned buffer will adopt the {@link #readerOffset()} of this buffer, and have its {@link #writerOffset()}
+     * and {@link #capacity()} both set to the equal to the write offset of this buffer.
+     * <p>
+     * The memory region in the returned buffer will become inaccessible through this buffer. This buffer will have its
+     * capacity reduced by the capacity of the returned buffer, and the read and write offsets of this buffer will both
+     * become zero, even though their position in memory remain unchanged.
+     * <p>
+     * Effectively, the following transformation takes place:
+     * <pre>{@code
+     *         This buffer:
+     *          +------------------------------------------+
+     *         0|   |r/o                  |w/o             |cap
+     *          +---+---------------------+----------------+
+     *         /   /                     / \               \
+     *        /   /                     /   \               \
+     *       /   /                     /     \               \
+     *      /   /                     /       \               \
+     *     /   /                     /         \               \
+     *    +---+---------------------+           +---------------+
+     *    |   |r/o                  |w/o & cap  |r/o & w/o      |cap
+     *    +---+---------------------+           +---------------+
+     *    Returned buffer.                      This buffer.
+     * }</pre>
+     * When the buffers are in this state, both of the bifurcated parts retain an atomic reference count on the
+     * underlying memory. This means that shared underlying memory will not be deallocated or returned to a pool, until
+     * all of the bifurcated parts have been closed.
+     * <p>
+     * Composite buffers have it a little easier, in that at most only one of the constituent buffers will actually be
+     * bifurcated. If the split point lands perfectly between two constituent buffers, then a composite buffer can
+     * simply split its internal array in two.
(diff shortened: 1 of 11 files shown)


src/main/java/io/netty/buffer/api/Buf.java
```diff
@@ -373,4 +373,48 @@ default ByteCursor openReverseCursor() {
...
+     *         /   /                     / \               \
+     *        /   /                     /   \               \
+     *       /   /                     /     \               \
+     *      /   /                     /       \               \
+     *     /   /                     /         \               \
+     *    +---+---------------------+           +---------------+
+     *    |   |r/o                  |w/o & cap  |r/o & w/o      |cap
+     *    +---+---------------------+           +---------------+
+     *    Returned buffer.                      This buffer.
+     * }</pre>
```
normanmaurer, line 406: <3 ascii art :)

src/main/java/io/netty/buffer/api/CompositeBuf.java
```diff
@@ -129,6 +130,7 @@ public String toString() {
 
     @Override
     public Buf order(ByteOrder order) {
+        this.order = order;
```
normanmaurer: nit: should we check that the order is actually changed and if not just return early ?

src/main/java/io/netty/buffer/api/Drop.java
```diff
@@ -15,16 +15,13 @@
...
-
 /**
  * The Drop interface is used by {@link Rc} instances to implement their resource disposal mechanics. The {@link
  * #drop(Object)} method will be called by the Rc when their last reference is closed.
  *
  * @param <T>
  */
-@FunctionalInterface
-public interface Drop<T> extends Consumer<T> {
+public interface Drop<T> {
```
normanmaurer, line 25: I guess we could still make this a Functional interface ?

src/main/java/io/netty/buffer/api/Drop.java
```diff
@@ -37,7 +34,6 @@
      *
      * @param obj The new Rc instance with the new owner.
      */
-    @Override
-    default void accept(T obj) {
+    default void reconnect(T obj) {
```
normanmaurer: maybe `attach` ?

src/main/java/io/netty/buffer/api/RcSupport.java
```diff
@@ -114,6 +117,12 @@ public int countBorrows() {
         return drop;
     }
 
+    protected Drop<T> unsafeExchangeDrop(Drop<T> replacement) {
+        Objects.requireNonNull(replacement, "Replacement drop cannot be null.");
+        drop = replacement;
```
normanmaurer: nit: you can merge both lines above.

src/main/java/io/netty/buffer/api/memseg/BifurcatedDrop.java
```diff
@@ -0,0 +1,61 @@
...
+        count = 2; // These are created by buffer bifurcation, so we initially have 2 references to this drop.
+    }
+
+    public synchronized void increment() {
+        checkValidState();
+        count++;
+    }
+
+    @Override
+    public synchronized void drop(T obj) {
```
normanmaurer, line 53: Couldn't we just use an `AtomicInteger` for the count and so remove the whole `synchronized` here ?

Usually its a bad idea to call any other methods that we don't know what exactly they do while holding a monitor. For example we don't know what the delegate exactly does which may be time-consuming etc. So if we really need a `synchronized` we should only hold it during acting on the count.

normanmaurer (changes requested):
Generally speaking I like the idea... Just left a few comments on the implementation itself

chrisvest:
@normanmaurer I've addressed your comments

src/main/java/io/netty/buffer/api/memseg/BifurcatedDrop.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+package io.netty.buffer.api.memseg;
+
+import io.netty.buffer.api.Drop;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+
+class BifurcatedDrop<T> implements Drop<T> {
+    private static final VarHandle COUNT;
```
normanmaurer, line 24: @chrisvest you use a varHandle to reduce allocations and memory overhead right ?

chrisvest: Yeah, don't see a need to have the counter in its own object. `AtomicIntegers` are good if you want to share them across objects.

normanmaurer (approved):

PR #17, opened 2020-12-14
Add BufHolder and BufRef helper classes

chrisvest:
Motivation:
There are many use cases where other objects will have fields that are buffers.
Since buffers are reference counted, their life cycle needs to be managed carefully.

Modification:
Add the abstract BufHolder, and the concrete sub-class BufRef, as neat building blocks for building other classes that contain field references to buffers.

The behaviours of closed/sent buffers have also been specified in tests, and tightened up in the code.

Result:
It is now easier to create classes/objects that wrap buffers.

### DIFF
--- a/src/main/java/io/netty/buffer/api/Allocator.java
+++ b/src/main/java/io/netty/buffer/api/Allocator.java
@@ -123,13 +123,22 @@ public interface Allocator extends AutoCloseable {
      * @param extension The buffer to extend the composite buffer with.
      */
     static void extend(Buf composite, Buf extension) {
-        if (composite.getClass() != CompositeBuf.class) {
+        if (!isComposite(composite)) {
             throw new IllegalArgumentException(
                     "Expected the first buffer to be a composite buffer, " +
                     "but it is a " + composite.getClass() + " buffer: " + composite + '.');
         }
         CompositeBuf buf = (CompositeBuf) composite;
         buf.extendWith(extension);
+    }
+
+    /**
+     * Check if the given buffer is a {@linkplain #compose(Buf...) composite} buffer or not.
+     * @param composite The buffer to check.
+     * @return {@code true} if the given buffer was created with {@link #compose(Buf...)}, {@code false} otherwise.
+     */
+    static boolean isComposite(Buf composite) {
+        return composite.getClass() == CompositeBuf.class;
     }
 
     /**
--- a/src/main/java/io/netty/buffer/api/BufHolder.java
+++ b/src/main/java/io/netty/buffer/api/BufHolder.java
@@ -86,12 +86,7 @@ public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
     @Override
     public Send<T> send() {
         var send = buf.send();
-        return new Send<T>() {
-            @Override
-            public T receive() {
-                return BufHolder.this.receive(send.receive());
-            }
-        };
+        return () -> receive(send.receive());
     }
 
     /**
@@ -116,7 +111,7 @@ public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
      *
      * @param newBuf The new {@link Buf} instance that is replacing the currently held buffer.
      */
-    protected void replace(Buf newBuf) {
+    protected final void replaceBuf(Buf newBuf) {
         try (var ignore = buf) {
             buf = newBuf.acquire();
         }
@@ -134,7 +129,7 @@ public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
      *
      * @param send The new {@link Buf} instance that is replacing the currently held buffer.
      */
-    protected void replace(Send<Buf> send) {
+    protected final void replaceBuf(Send<Buf> send) {
         try (var ignore = buf) {
             buf = send.receive();
         }
@@ -152,7 +147,7 @@ public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
      *
      * @param newBuf The new {@link Buf} instance that is replacing the currently held buffer.
      */
-    protected void replaceVolatile(Buf newBuf) {
+    protected final void replaceBufVolatile(Buf newBuf) {
         var prev = (Buf) BUF.getAndSet(this, newBuf.acquire());
         prev.close();
     }
@@ -167,9 +162,9 @@ public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
      * <p>
      * The buffer assignment is performed using a volatile store.
      *
-     * @param send The new {@link Buf} instance that is replacing the currently held buffer.
+     * @param send The {@link Send} with the new {@link Buf} instance that is replacing the currently held buffer.
      */
-    protected void replaceVolatile(Send<Buf> send) {
+    protected final void replaceBufVolatile(Send<Buf> send) {
         var prev = (Buf) BUF.getAndSet(this, send.receive());
         prev.close();
     }
@@ -181,7 +176,7 @@ public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
      *
      * @return The {@link Buf} instance being held by this {@linkplain T buffer holder}.
      */
-    protected Buf getBuf() {
+    protected final Buf getBuf() {
         return buf;
     }
 
@@ -192,7 +187,7 @@ public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
      *
      * @return The {@link Buf} instance being held by this {@linkplain T buffer holder}.
      */
-    protected Buf getBufVolatile() {
+    protected final Buf getBufVolatile() {
         return (Buf) BUF.getVolatile(this);
     }
 }
--- a/src/main/java/io/netty/buffer/api/BufHolder.java
+++ b/src/main/java/io/netty/buffer/api/BufHolder.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2020 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer.api;
+
+import java.lang.invoke.VarHandle;
+import java.util.Objects;
+
+import static io.netty.buffer.api.Statics.findVarHandle;
+import static java.lang.invoke.MethodHandles.lookup;
+
+/**
+ * The {@link BufHolder} is an abstract class that simplifies the implementation of objects that themselves contain
+ * a {@link Buf} instance.
+ * <p>
+ * The {@link BufHolder} can only hold on to a single buffer, so objects and classes that need to hold on to multiple
+ * buffers will have to do their implementation from scratch, though they can use the code of the {@link BufHolder} as
+ * inspiration.
+ * <p>
+ * If you just want an object that is a reference to a buffer, then the {@link BufRef} can be used for that purpose.
+ * If you have an advanced use case where you wish to implement {@link Rc}, and tightly control lifetimes, then
+ * {@link RcSupport} can be of help.
+ *
+ * @param <T> The concrete {@link BufHolder} type.
+ */
+public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
+    private static final VarHandle BUF = findVarHandle(lookup(), BufHolder.class, "buf", Buf.class);
+    private Buf buf;
+
+    /**
+     * Create a new {@link BufHolder} to hold the given {@linkplain Buf buffer}.
+     * <p>
+     * <strong>Note:</strong> this increases the reference count of the given buffer.
+     *
+     * @param buf The {@linkplain Buf buffer} to be held by this holder.
+     */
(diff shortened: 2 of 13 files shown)


src/main/java/io/netty/buffer/api/Allocator.java
```diff
@@ -132,6 +132,15 @@ static void extend(Buf composite, Buf extension) {
         buf.extendWith(extension);
     }
 
+    /**
+     * Check if the given buffer is a {@linkplain #compose(Buf...) composite} buffer or not.
+     * @param composite The buffer to check.
+     * @return {@code true} if the given buffer was created with {@link #compose(Buf...)}, {@code false} otherwise.
+     */
+    static boolean isComposite(Buf composite) {
+        return composite.getClass() == CompositeBuf.class;
```
normanmaurer, line 141: Just a question... why not just `instanceof ` ?

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+ * <p>
+ * If you just want an object that is a reference to a buffer, then the {@link BufRef} can be used for that purpose.
+ * If you have an advanced use case where you wish to implement {@link Rc}, and tightly control lifetimes, then
+ * {@link RcSupport} can be of help.
+ *
+ * @param <T> The concrete {@link BufHolder} type.
+ */
+public abstract class BufHolder<T extends BufHolder<T>> implements Rc<T> {
+    private static final VarHandle BUF = findVarHandle(lookup(), BufHolder.class, "buf", Buf.class);
+    private Buf buf;
```
normanmaurer, line 40: Does this need to be `volatile` ? We use `getAndSet(...)` below so I would suspect it must be volatile.

src/main/java/io/netty/buffer/api/BufRef.java
```diff
@@ -0,0 +1,68 @@
...
+public final class BufRef extends BufHolder<BufRef> {
+    /**
+     * Create a reference to the given {@linkplain Buf buffer}.
+     * This increments the reference count of the buffer.
+     *
+     * @param buf The buffer to reference.
+     */
+    public BufRef(Buf buf) {
+        super(buf);
+        VarHandle.fullFence();
```
normanmaurer, line 33: can you add a comment why this is called ?

src/main/java/io/netty/buffer/api/BufRef.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Create a reference that holds the exclusive ownership of the sent buffer.
+     *
+     * @param send The {@linkplain Send sent} buffer to take ownership of.
+     */
+    public BufRef(Send<Buf> send) {
+        super(send);
+        VarHandle.fullFence();
```
normanmaurer, line 44: can you add a comment why this is called ?

src/main/java/io/netty/buffer/api/memseg/BifurcatedDrop.java
```diff
@@ -50,7 +50,7 @@ void increment() {
     }
 
     @Override
-    public synchronized void drop(T obj) {
+    public synchronized void drop(MemSegBuf buf) {
```
normanmaurer: Not really related to this PR but I think we should try to reduce the `synchronised` scope. Who knows that the delegate does...

normanmaurer (changes requested):

src/main/java/io/netty/buffer/api/Allocator.java
chrisvest: There are no sub-classes of CompositeBuf.

src/main/java/io/netty/buffer/api/BufHolder.java
chrisvest: I've added methods to expose both plain and volatile access modes for this field.

src/main/java/io/netty/buffer/api/memseg/BifurcatedDrop.java
chrisvest: Forgot to remove the synchronised after changing to using atomic ops.

chrisvest:
@normanmaurer Comments addressed.

src/main/java/io/netty/buffer/api/Allocator.java
normanmaurer: sure... But I wondered if this is faster then instanceof ?

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Create a new {@link BufHolder} to hold the given {@linkplain Buf buffer}.
+     * <p>
+     * <strong>Note:</strong> this increases the reference count of the given buffer.
```
normanmaurer, line 45: interesting... So this is a bit different to what we do in netty atm. I wonder if this is a good idea as usually people do something like this:

```java
DefaultByteBufHolder holder = new DefaultByteBufHolder(Unpooled.buffer());
```

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Access the held {@link Buf} instance.
+     * <p>
+     * The access is performed using a volatile load.
+     *
+     * @return The {@link Buf} instance being held by this {@linkplain T buffer holder}.
+     */
+    protected Buf getBufVolatile() {
```
normanmaurer: final ? I don't see why anything would ever override this

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Access the held {@link Buf} instance.
+     * <p>
+     * The access is performed using a plain load.
+     *
+     * @return The {@link Buf} instance being held by this {@linkplain T buffer holder}.
+     */
+    protected Buf getBuf() {
```
normanmaurer: final ? I don't see why anything would ever override this

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+     * This method is protected to permit advanced use cases of {@link BufHolder} sub-class implementations.
+     * <p>
+     * <strong>Note:</strong> this method decreases the reference count of the current buffer,
+     * and takes exclusive ownership of the sent buffer.
+     * <p>
+     * The buffer assignment is performed using a volatile store.
+     *
+     * @param send The new {@link Buf} instance that is replacing the currently held buffer.
+     */
+    protected void replaceVolatile(Send<Buf> send) {
```
normanmaurer: final ? I don't see why anything would ever override this

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+     * This method is protected to permit advanced use cases of {@link BufHolder} sub-class implementations.
+     * <p>
+     * <strong>Note:</strong> this method decreases the reference count of the current buffer,
+     * and increases the reference count of the new buffer.
+     * <p>
+     * The buffer assignment is performed using a volatile store.
+     *
+     * @param newBuf The new {@link Buf} instance that is replacing the currently held buffer.
+     */
+    protected void replaceVolatile(Buf newBuf) {
```
normanmaurer: final ? I don't see why anything would ever override this

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+     * This method is protected to permit advanced use cases of {@link BufHolder} sub-class implementations.
+     * <p>
+     * <strong>Note:</strong> this method decreases the reference count of the current buffer,
+     * and takes exclusive ownership of the sent buffer.
+     * <p>
+     * The buffer assignment is performed using a plain store.
+     *
+     * @param send The new {@link Buf} instance that is replacing the currently held buffer.
+     */
+    protected void replace(Send<Buf> send) {
```
normanmaurer: final ? I don't see why anything would ever override this

src/main/java/io/netty/buffer/api/BufHolder.java
```diff
@@ -0,0 +1,198 @@
...
+     * This method is protected to permit advanced use cases of {@link BufHolder} sub-class implementations.
+     * <p>
+     * <strong>Note:</strong> this method decreases the reference count of the current buffer,
+     * and increases the reference count of the new buffer.
+     * <p>
+     * The buffer assignment is performed using a plain store.
+     *
+     * @param newBuf The new {@link Buf} instance that is replacing the currently held buffer.
+     */
+    protected void replace(Buf newBuf) {
```
normanmaurer: final ? I don't see why anything would ever override this

src/main/java/io/netty/buffer/api/BufRef.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public void replaceVolatile(Buf newBuf) {
+        super.replaceVolatile(newBuf);
+    }
+
+    @Override
+    public void replaceVolatile(Send<Buf> send) {
+        super.replaceVolatile(send);
```
normanmaurer: I just noticed because of this overrides it may not be possible to mark the methods as final... :/

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -389,28 +415,44 @@ public int getUnsignedByte(int roff) {
 
     @Override
     public Buf writeByte(byte value) {
-        setByteAtOffset(seg, woff, value);
-        woff += Byte.BYTES;
-        return this;
+        try {
+            setByteAtOffset(seg, woff, value);
+            woff += Byte.BYTES;
```
normanmaurer, line 420: do we need to guard against overflow ?

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -389,28 +415,44 @@ public int getUnsignedByte(int roff) {
...
     }
 
     @Override
     public Buf writeUnsignedByte(int value) {
-        setByteAtOffset(seg, woff, (byte) (value & 0xFF));
-        woff += Byte.BYTES;
-        return this;
+        try {
+            setByteAtOffset(seg, woff, (byte) (value & 0xFF));
+            woff += Byte.BYTES;
```
normanmaurer, line 441: do we need to guard against overflow ?

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -429,15 +471,23 @@ public char getChar(int roff) {
 
     @Override
     public Buf writeChar(char value) {
-        setCharAtOffset(seg, woff, order, value);
-        woff += 2;
-        return this;
+        try {
+            setCharAtOffset(seg, woff, order, value);
+            woff += 2;
```
normanmaurer, line 476: do we need to guard against overflow ?

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -470,28 +520,44 @@ public int getUnsignedShort(int roff) {
 
     @Override
     public Buf writeShort(short value) {
-        setShortAtOffset(seg, woff, order, value);
-        woff += Short.BYTES;
-        return this;
+        try {
+            setShortAtOffset(seg, woff, order, value);
+            woff += Short.BYTES;
```
normanmaurer, line 525: do we need to guard against overflow ?

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -470,28 +520,44 @@ public int getUnsignedShort(int roff) {
...
     }
 
     @Override
     public Buf writeUnsignedShort(int value) {
-        setShortAtOffset(seg, woff, order, (short) (value & 0xFFFF));
-        woff += Short.BYTES;
-        return this;
+        try {
+            setShortAtOffset(seg, woff, order, (short) (value & 0xFFFF));
+            woff += Short.BYTES;
```
normanmaurer, line 546: do we need to guard against overflow ?

Same question everywhere I guess

normanmaurer (changes requested):

chrisvest:
@normanmaurer We don't need to guard against overflow, because we won't allow the construction of a buffer bigger than `Integer.MAX_VALUE - 8`.

src/main/java/io/netty/buffer/api/BufRef.java
chrisvest: We could come up with different names that don't overlap. I can see the appeal of having those methods be final.

src/main/java/io/netty/buffer/api/BufHolder.java
chrisvest: Every allocation site should be paired with a deallocation site. This should ideally follow from the structure of the code, e.g. with try-with-resources. For cases where this is too rigid to express what one want, you use send.

src/main/java/io/netty/buffer/api/Allocator.java
chrisvest: It's certainly not slower, because it boils down to a pointer equality check, while `instanceof` may have to traverse the type hierarchy.

normanmaurer: ok... was just wondering as I usually just use instanceof.

src/main/java/io/netty/buffer/api/BufHolder.java
normanmaurer: got it... Maybe I just need to get used to it and see some real-world-code translated to use the new stuff

normanmaurer (approved):

PR #19, opened 2020-12-17
Add a benchmark that explore the overhead of always attaching a cleaner to buffers

chrisvest:
Looks like the overhead is not too bad, so I think we can just always do that:

```
Benchmark                       (workload)  Mode  Cnt  Score   Error  Units
explicitPooledClose                  light  avgt  150  1,094 ± 0,017  us/op
pooledWithCleanerExplicitClose       light  avgt  150  1,181 ± 0,009  us/op
```

This PR is baselined on #18


chrisvest:
And for the heavy workload, similar story:

```
Benchmark                       (workload)  Mode  Cnt   Score   Error  Units
explicitPooledClose                  heavy  avgt  150  11,860 ± 0,175  us/op
pooledWithCleanerExplicitClose       heavy  avgt  150  14,064 ± 0,432  us/op
```

For the light workload, the overhead is ~8%, while for the heavy workload, we see a ~18.6% increase.

normanmaurer (approved):

normanmaurer:
I think we could just always attach a cleaner until proven otherwise.

chrisvest:
Alright, I'll move ahead make cleaners attached by default. As we discussed elsewhere, I think the perf. hit on the heavy workload can be decreased and get closer to that of the light workload.

PR #24, opened 2021-01-11
Add support for iterating underlying buffer components

chrisvest:
Motivation:
It's desirable to be able to access the contents of a Buf via an array or a ByteBuffer.
However, we would also like to have a unified API that works for both composite and non-composite buffers.
Even for nested composite buffers.

Modification:
Add a forEachReadable method, which uses internal iteration to process all buffer components.
The internal iteration allows us to hide any nesting of composite buffers.
The consumer in the internal iteration is presented with a Component object, which exposes the contents in various ways.
The data is exposed from the Component via methods, such that anything that is expensive to create, will not have to be paid for unless it is used.
This mechanism also let us avoid any allocation unnecessary allocation; the ByteBuffers and arrays will necessarily have to be allocated, but the consumer may or may not need allocation depending on how it's implemented, and the component objects do not need to be allocated, because the non-composite buffers can directly implement the Component interface.

Result:
It's now possible to access the contents of Buf instances as arrays or ByteBuffers, without having to copy the data.


chrisvest:
We also need an equivalent `forEachWritable`, I think. Not sure if it's technically possible, though.

src/main/java/io/netty/buffer/api/Component.java
```diff
@@ -0,0 +1,57 @@
...
+     * Check if this component is backed by a cached byte array than can be accessed cheaply.
+     *
+     * @return {@code true} if {@link #array()} is a cheap operation, otherwise {@code false}.
+     */
+    boolean hasCachedArray();
+
+    /**
+     * Get a byte array of the contents of this component.
+     * <p>
+     * <strong>Note</strong> that the array is meant to be read-only. It may either be a direct reference to the
```
normanmaurer: nit: consider also adding a sentence that if you write to the array it is "undefined" behaviour.

src/main/java/io/netty/buffer/api/Component.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * A view onto the buffer component being processed in a given iteration of {@link Buf#forEachReadable(Consumer)}.
+ * <p>
+ * Instances of this interface are allowed to be mutable behind the scenes, and the data is only guaranteed to be
+ * consistent within the given iteration.
+ */
+public interface Component {
+
+    /**
+     * Check if this component is backed by a cached byte array than can be accessed cheaply.
```
normanmaurer: I think I would also state the the contract is the same as `array` in the sense that this is not meant to be written too.

src/main/java/io/netty/buffer/api/Component.java
```diff
@@ -0,0 +1,57 @@
...
+    byte[] array();
+
+    /**
+     * Give the native memory address backing this buffer, or return 0 if this buffer has no native memory address.
+     * @return The native memory address, if any, otherwise 0.
+     */
+    long nativeAddress();
+
+    /**
+     * Build a {@link ByteBuffer} instance for this memory component.
```
normanmaurer: Should we also note that it build a "read only" instance

src/main/java/io/netty/buffer/api/Component.java
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @return A byte array of the contents of this component.
+     */
+    byte[] array();
+
+    /**
+     * Give the native memory address backing this buffer, or return 0 if this buffer has no native memory address.
+     * @return The native memory address, if any, otherwise 0.
+     */
+    long nativeAddress();
```
normanmaurer: Consider adding a note that this should not be used to write to the buffer ?

src/main/java/io/netty/buffer/api/Component.java
```diff
@@ -0,0 +1,57 @@
...
+import java.nio.ByteBuffer;
+import java.util.function.Consumer;
+
+/**
+ * A view onto the buffer component being processed in a given iteration of {@link Buf#forEachReadable(Consumer)}.
+ * <p>
+ * Instances of this interface are allowed to be mutable behind the scenes, and the data is only guaranteed to be
+ * consistent within the given iteration.
+ */
+public interface Component {
```
normanmaurer: as this is used for read operations we may want to make it more clear that all the methods here return "read-only" things by adding it to the classname. Something like `ReadableComponent`, or `ReadOnlyComponent`

src/main/java/io/netty/buffer/api/CompositeBuf.java
```diff
@@ -700,6 +701,27 @@ public void compact() {
         writerOffset(woff - distance);
     }
 
+    @Override
+    public int componentCount() {
+        int sum = 0;
+        for (Buf buf : bufs) {
+            sum += buf.componentCount();
+        }
+        return sum;
```
normanmaurer, line 752: This could in theory overflow. I think we need to guard against it.

src/main/java/io/netty/buffer/api/CompositeBuf.java
```diff
@@ -700,6 +701,27 @@ public void compact() {
...
+    }
+
+    @Override
+    public int forEachReadable(Consumer<Component> consumer) {
+        checkReadBounds(readerOffset(), Math.max(1, readableBytes()));
+        int visited = 0;
+        for (Buf buf : bufs) {
+            if (buf.readableBytes() > 0) {
+                visited += buf.forEachReadable(consumer);
```
normanmaurer: This could in theory overflow. I think we need to guard against it.

normanmaurer (changes requested):

src/main/java/io/netty/buffer/api/Component.java
chrisvest: It currently doesn't always do that (currently only if the Buf instance itself is read-only), but thinking about it, perhaps it should.

src/main/java/io/netty/buffer/api/Component.java
chrisvest: Makes sense, because I've realised we also need a version for writable components.

src/main/java/io/netty/buffer/api/CompositeBuf.java
chrisvest: It could, since we don't prevent creating zero-sized buffers, or prevent composite buffers from containing themselves.

Throwing an exception at this point would be inconvenient, but returning a `long` could also overflow after a few hundred years. Not sure which approach is better.

chrisvest: Hmm… I think we can still get away with checking this up front, because only owned composite buffers can extended, and the only way for someone to extend an owned composite buffer that is a constituent component of another composite buffer, is to hold a reference to it after a `close()` call. Doing that is obviously undefined. Then, we can also filter away buffers that are empty, since there's no way for them to contribute to a composite buffer.

With all that in place, we are safe against overflow because now the number of components can not be greater than the number of bytes, _except_ when a composite buffer contains itself, in which case the number of components is infinite.

src/main/java/io/netty/buffer/api/Component.java
chrisvest: Thinking about this some more… We would probably like to have things like `IovArray` be able to abstract over these components, both for reading and writing. So a single component type that works for both is still desirable.

normanmaurer: I think we could do something similar to `Channel` which extends `WritableByteChannel` and `ReadableByteChannel`. So I would prefer to have a specific type for reading and one for writing.

src/main/java/io/netty/buffer/api/CompositeBuf.java
normanmaurer: > Hmm… I think we can still get away with checking this up front, because only owned composite buffers can extended, and the only way for someone to extend an owned composite buffer that is a constituent component of another composite buffer, is to hold a reference to it after a close() call. Doing that is obviously undefined. Then, we can also filter away buffers that are empty, since there's no way for them to contribute to a composite buffer.

Hmm... How would this guard up fron if the `Buf` implementation may be custom ?

> With all that in place, we are safe against overflow because now the number of components can not be greater than the number of bytes, except when a composite buffer contains itself, in which case the number of components is infinite.

We should ensure you can not add "yourself" to a `composite buffer`

src/main/java/io/netty/buffer/api/Component.java
normanmaurer: I think we should ...

src/main/java/io/netty/buffer/api/CompositeBuf.java
chrisvest: > Hmm... How would this guard up fron if the `Buf` implementation may be custom ?

This is in the spec of the `Buf` interface, and for composite buffers. Buffers must be owned to change capacity; buffers that become part of a composite buffer have their ref counts incremented (preventing ownership if there are other references); if you relinquish your reference to a buffer, you're not allowed to touch it anymore.

How the buffers are implemented won't matter as long as they are compliant.

> We should ensure you can not add "yourself" to a `composite buffer`

We'll have to unwrap any nesting before we can make this check. I have a card for this already.

chrisvest:
@normanmaurer How about this?

src/main/java/io/netty/buffer/api/Buf.java
```diff
@@ -482,4 +487,117 @@ default void ensureWritable(int size) {
...
+     * <p>
+     * <strong>Note</strong> that the {@link ReadableComponent} instance passed to the consumer could be reused for
+     * multiple calls, so the data must be extracted from the component in the context of the iteration.
+     * <p>
+     * The {@link ByteBuffer} instances obtained from the component, share life time with that internal component.
+     * This means they can be accessed as long as the internal memory store remain unchanged. Methods that may cause
+     * such changes, are any method that requires the buffer to be {@linkplain #isOwned() owned}.
+     * <p>
+     * The best way to ensure this doesn't cause any trouble, is to use the buffers directly as part of the iteration,
+     * or immediately after the iteration.
```
normanmaurer: should we make this something like ?:

```
....or immediately after the iteration while we are still in the scope of the method that triggered the iteration.
```

src/main/java/io/netty/buffer/api/Buf.java
```diff
@@ -482,4 +487,117 @@ default void ensureWritable(int size) {
...
+     * <p>
+     * <strong>Note</strong> that the {@link WritableComponent} instance passed to the consumer could be reused for
+     * multiple calls, so the data must be extracted from the component in the context of the iteration.
+     * <p>
+     * The {@link ByteBuffer} instances obtained from the component, share life time with that internal component.
+     * This means they can be accessed as long as the internal memory store remain unchanged. Methods that may cause
+     * such changes, are any method that requires the buffer to be {@linkplain #isOwned() owned}.
+     * <p>
+     * The best way to ensure this doesn't cause any trouble, is to use the buffers directly as part of the iteration,
+     * or immediately after the iteration.
```
normanmaurer: Same comment as above.

src/main/java/io/netty/buffer/api/ComponentProcessor.java
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * This interface contain a collection of APIs used in the {@link Buf#forEachReadable(int, OfReadable)} and
+ * {@link Buf#forEachWritable(int, OfWritable)} methods.
+ */
+public interface ComponentProcessor {
+    /**
+     * A processor of {@linkplain ReadableComponent readable components}.
+     */
+    @FunctionalInterface
+    interface OfReadable extends ComponentProcessor {
```
normanmaurer: not a big fan of the name... Maybe its just because it starts with `Of...`..

src/main/java/io/netty/buffer/api/ComponentProcessor.java
```diff
@@ -0,0 +1,169 @@
...
+         * {@code false} to stop the iteration early.
+         */
+        boolean process(int index, ReadableComponent component);
+    }
+
+    /**
+     * A processor of {@linkplain WritableComponent writable components}.
+     */
+    @FunctionalInterface
+    interface OfWritable extends ComponentProcessor {
```
normanmaurer: same comment as above... Maybe just call it `Writable`

src/main/java/io/netty/buffer/api/ComponentProcessor.java
```diff
@@ -0,0 +1,169 @@
...
+         */
+        long readableNativeAddress();
+
+        /**
+         * Get a {@link ByteBuffer} instance for this memory component.
+         * <p>
+         * <strong>Note</strong> that the {@link ByteBuffer} is read-only, to prevent write accesses to the memory,
+         * when the buffer component is obtained through {@link Buf#forEachReadable(int, OfReadable)}.
+         *
+         * @return A new {@link ByteBuffer} for this memory component.
```
normanmaurer: maybe just add something like `... which has its own position / limit`. I know this is kind of already hinted with `New` but maybe it makes it easier for people to understand the semantics.

src/main/java/io/netty/buffer/api/ComponentProcessor.java
```diff
@@ -0,0 +1,169 @@
...
+         */
+        long writableNativeAddress();
+
+        /**
+         * Get a {@link ByteBuffer} instance for this memory component, which can be used for modifying the buffer
+         * contents.
+         *
+         * @return A new {@link ByteBuffer} for this memory component.
+         */
+        ByteBuffer writableBuffer();
```
normanmaurer, line 168: Same comment as above.

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
```diff
@@ -58,22 +61,25 @@
 
     private final AllocatorControl alloc;
     private final boolean isSendable;
+    private final int baseOffset; // TODO remove this when JDK bug is fixed (slices of heap buffers)
```
normanmaurer: can we report this bug via the mailing list and add the link here ?

normanmaurer (approved):
All in all this looks great... just some nits.

src/main/java/io/netty/buffer/api/memseg/MemSegBuf.java
chrisvest: It's already reported. Can add a mailing list link.

src/main/java/io/netty/buffer/api/ComponentProcessor.java
chrisvest: Same pattern as how `Spliterator` specialises.

normanmaurer: wasn't aware of this... so I guess it's fine. Matter of taste :)

PR #30, opened 2021-02-15
Writeup of rationale behind the buffer API design

chrisvest:
This adds a write-up of the rationale behind the design of the new proposed buffer API, and also draws attention to some of the major changes compared to the `ByteBuf` API.

### DIFF
--- a/RATIONALE.adoc
+++ b/RATIONALE.adoc
@@ -219,17 +219,17 @@ This is to avoid the pains currently observed where we code that branches on whe
 Being able to unify these code paths will help with maintainability.
 
 Buffers need to know their allocators, in order to implement `ensureWritable()`, and the same is true for composite buffers.
-That’s why the method to compose buffers is placed on the `BufferAllocator` interface:
+That’s why the method to compose buffers takes a `BufferAllocator` as a first argument:
 
 [source,java]
 ----
 try (Buffer x = allocator.allocate(128);
      Buffer y = allocator.allocate(128)) {
-    return allocator.compose(x, y);
+    return Buffer.compose(allocator, x, y);
 }
 ----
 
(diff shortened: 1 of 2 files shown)


chrisvest:
This needs to mention that frameworks that build upon this, and don't explicitly manage buffer life times, should also be supported.

chrisvest:
See #41 for slides with examples and general introduction of the API.

jameskleeh:
@chrisvest I have some concerns regarding the changes to composite buffers. 

In Micronaut we currently need a way to tell if a buffer is composite or not via https://github.com/micronaut-projects/micronaut-core/blob/2.4.x/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBuffer.java#L187

We also need a way to release a section of a composite buffer. This need and the current shortcomings of the current file upload API are described here https://github.com/netty/netty/issues/8655. See https://github.com/micronaut-projects/micronaut-core/blob/2.4.x/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpDataReference.java#L221-L243

If this isn't the right forum to bring up these issues or you would like me to bring them up somewhere else please let me know

chrisvest:
Hi @jameskleeh, opening separate issues is better for keeping track of things, but I can answer the questions you have here.

We have a static `Buffer.isComposite` method for determining if a buffer is composite or not.

You can use `bifurcate`, assuming the offsets line up correctly, to break a composite buffer into two. Then you can close one of the two parts. We could also add methods that make detaching/removing components more explicit and easier to express. Please open an issue about that and describe the use case and what you need.

chrisvest:
Merging this one. We can do future discussion as GitHub issues.

PR #32, opened 2021-02-24
Make the incubating buffers exposable as ByteBuf

chrisvest:
Motivation:
This makes it possible to use the new buffer API in Netty as is.

Modification:
Make the MemSegBuffer implementation class implement ByteBufConvertible and ReferenceCounted.
The produced ByteBuf instance delegates all calls to the underlying Buffer instance as faithfully as possible.
One area where the two deviates, is that it's not possible to create non-retained duplicates and slices with the new buffer API.

Result:
It is now possible to use the new buffer API on both client and server side.
The Echo* examples demonstrate this, and the EchoIT proves it with a test.
The API is used more directly on the client side, since the server-side allocator in Netty does not know how to allocate buffers with the incubating API.

This is the companion PR to https://github.com/netty/netty/pull/11036


chrisvest:
GitHub Actions is having an outage. Build passes locally.

PR #37, opened 2021-03-18
Add a ByteBuffer based implementation of Buffer

chrisvest:
Motivation:
We need a new implementation of our new API that supports Java 11, since that is what Netty 5 will most likely baseline on.
We also need an implementation that does not rely on Unsafe.
This leaves us with ByteBuffer as the underlying currency of memory.

Modification:
- Add a NioBuffer implementation and associated supporting classes.
- The entry-point for this is a new MemoryManagers API, which is used to pick the implementation and provide the on-/off-heap MemoryManager implementations.
- Add a mechanism to configure/override which MemoryManagers implementation to use.
- The MemoryManagers implementations are service-loadable, so new ones can be discovered at runtime.
- The existing MemorySegment based implementation also get a MemoryManagers implementation.
- Expand the BufferTest to include all combinations of all implementations. We now run 360.000 tests in BufferTest.
- Some common infrastructure, like ArcDrop, is moved to its own package.
- Add a module-info.java to control the service loading, and the visibility in the various packages.
- Some pom.xml file updates to support our now module based project.

Result:
We have an implementation that should work on Java 11, but we currently don't build or test on 11.
More work needs to happen before that is a reality.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
   <properties>
     <javaModuleName>io.netty.incubator.buffer</javaModuleName>
     <netty.version>5.0.0.Final-SNAPSHOT</netty.version>
-    <netty.build.version>29-SNAPSHOT</netty.build.version>
+    <netty.build.version>29</netty.build.version>
     <java.version>16</java.version>
     <junit.version>5.7.0</junit.version>
     <surefire.version>3.0.0-M5</surefire.version>
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
   <properties>
     <javaModuleName>io.netty.incubator.buffer</javaModuleName>
     <netty.version>5.0.0.Final-SNAPSHOT</netty.version>
-    <netty.build.version>28</netty.build.version>
+    <netty.build.version>29-SNAPSHOT</netty.build.version>
     <java.version>16</java.version>
     <junit.version>5.7.0</junit.version>
     <surefire.version>3.0.0-M5</surefire.version>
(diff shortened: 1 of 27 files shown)


chrisvest:
Draft for the time being because I think we need some Netty changes before we can access the `AbstractByteBufTest` at test-compile time.

pom.xml
```diff
@@ -69,7 +69,7 @@
   <properties>
     <javaModuleName>io.netty.incubator.buffer</javaModuleName>
     <netty.version>5.0.0.Final-SNAPSHOT</netty.version>
-    <netty.build.version>28</netty.build.version>
+    <netty.build.version>29-SNAPSHOT</netty.build.version>
```
normanmaurer: @chrisvest we can cut a release if you like

normanmaurer (approved):

pom.xml
chrisvest: @normanmaurer Yes, let's do that. I just now managed to verify that I don't need any other changes to netty-build.

pom.xml
```diff
@@ -123,14 +123,13 @@
           <compilerArgs>
             <arg>--add-modules</arg>
             <arg>jdk.incubator.foreign</arg>
+
+            <!--
+             These two are really only needed for test-compile, but the maven-compiler-plugin cannot express that
+             -->
+            <arg>--patch-module</arg>
+            <arg>io.netty.buffer=${io.netty:netty-buffer:test-jar:tests}</arg>
```
chrisvest, line 131: This has to be placed in `compilerArgs` instead of the various `testCompiler*` configurations, because the latter cannot correctly handle space-separated argument-value pairs.

This is a `maven-compiler-plugin` limitation, that https://github.com/apache/maven-compiler-plugin/pull/27 will hopefully solve.

PR #38, opened 2021-03-24
Split modules experiment

chrisvest:
This seems to be a dead-end since Intellij can't run tests when modules are set up like this.


chrisvest:
Closing because this is just for show-casing; not meant to be merged.

PR #39, opened 2021-03-26
UnsafeBuffer implementation, based on sun.misc.Unsafe

chrisvest:
Motivation:
When Unsafe is available, we can supposedly do certain things faster than when it is not.

Modification:
Add a Buffer implementation that take advantage of sun.misc.Unsafe.
It has not yet been verified if this is faster in any way than, say the ByteBuffer implementation or the MemorySegment implementation.

Result:
Another Buffer implementation that can be used when Unsafe is available.


chrisvest:
Draft PR because there are still test failures and JVM crashes.

src/main/java/io/netty/buffer/api/unsafe/UnsafeMemoryManagers.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package io.netty.buffer.api.unsafe;
+
+import io.netty.buffer.api.MemoryManager;
+import io.netty.buffer.api.MemoryManagers;
+import io.netty.util.internal.PlatformDependent;
+
+public class UnsafeMemoryManagers implements MemoryManagers {
+    public UnsafeMemoryManagers() {
+        if (!PlatformDependent.hasUnsafe()) {
```
hyperxpro, line 24: I think this should go in the `static` block. WDYT?

src/main/java/module-info.java
```diff
@@ -32,5 +32,6 @@
 
     provides io.netty.buffer.api.MemoryManagers with
             io.netty.buffer.api.memseg.SegmentMemoryManagers,
-            io.netty.buffer.api.bytebuffer.ByteBufferMemoryManagers;
+            io.netty.buffer.api.bytebuffer.ByteBufferMemoryManagers,
+            io.netty.buffer.api.unsafe.UnsafeMemoryManagers;
 }
```
hyperxpro: Empty Line Missing.

src/main/java/io/netty/buffer/api/unsafe/UnsafeMemoryManagers.java
chrisvest: These instances are service loaded, so I'll have to check how that interacts… currently the provider will throw, so it can be handled when they are iterated.

chrisvest: The only difference this does is that the exception thrown from the provider will have an additional ExceptionInInitializerError cause level. So I don't think it makes sense to make that change.

PR #40, opened 2021-04-06
Use the memaccess+abi branch of panama-foreign

chrisvest:
Motivation:
The jextract branch is currently suffering from upstream compilation
errors.
And we don't need jextract - memaccess+abi is enough.

Modification:
Avoid checking out the jextract branch of panama by default, and instead
explicitly check out the foreign-memaccess+abi branch when we clone the
JDK.

Result:
Our build should now be able to proceed again.


chrisvest:
The OpenJDK build is fixed in this PR. Updating the memory segment implementation of the Buffer API will be done in another PR.

PR #44, opened 2021-04-23
More docs and examples of slice() and bifurcate()

chrisvest:
* Class-level documentation has been added to the `Buffer` interface, which describes when you would use `slice()` and when you would use `bifurcate()`.
* A `bifurcate(int)` overload method has been added, so it's possible to split a buffer at an offset that isn't the write offset.
* The `ByteToMessageDecoder` and its test has been ported to the new API, as an advanced example of how to use the new API.
* An `AlternativeMessageDecoder` and companion test, has also been implemented to show a slightly different way of implementing the same pattern. This implementation is easier to read because it has fewer complications and features inherited from the product version.
* The `ensureWritable` method can now grow the buffer in increments that are greater than what is strictly necessary to satisfy the immediate need. This can help amortise the cost of reallocation.
* A `BufferAdaptor` class has been added, which delegates all method calls to a delegate field. The idea is that the individual methods can be overridden in subclasses. This is used the `ByteToMessageDecoderTest`.

@njhill @ejona86 You had questions about slice vs. bifurcate, and what it would look like to have conditional ownership transfer. Please take a look at this and let me know what you think!


src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -101,6 +99,34 @@
...
+ *         be visible through the other.
+ *     </li>
+ * </ul>
+ *
+ * These differences means that slicing is mostly suitable for when you temporarily want to share a focused area of a
+ * buffer.
+ * Examples of this include doing IO, or decoding a bounded part of a larger message.
+ * On the other hand, bifurcate is suitable for when you want to hand over a region of a buffer to some other,
+ * perhaps unknown, piece of code, and relinquish your ownership of that buffer region in the process.
+ * Examples of include aggregating messages into an accumulator buffer, and sending messages down the pipeline for
```
ejona86: "Examples of include". Seems like a word was dropped or should be dropped.

src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -558,10 +594,64 @@ default void ensureWritable(int size) {
...
+        return bifurcate(writerOffset());
+    }
+
+    /**
+     * Split the buffer into two, at the given {@code splitOffset}.
+     * <p>
+     * The buffer must be in {@linkplain #isOwned() an owned state}, or an exception will be thrown.
+     * <p>
+     * The region of this buffer that precede the {@code splitOffset}, will be captured and returned in a new
+     * buffer, that will hold its own ownership of that region. This allows the returned buffer to be indepentently
```
ejona86: s/indepentently/independently/

src/test/java/io/netty/buffer/api/examples/bytetomessagedecoder/AlternativeMessageDecoder.java
```diff
@@ -0,0 +1,115 @@
...
+        this.allocator = Objects.requireNonNull(allocator, "BufferAllocator cannot be null.");
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof Buffer) {
+            try (Buffer input = (Buffer) msg) {
+                processRead(ctx, input);
+            }
+        } else if (msg instanceof Send && ((Send<?>) msg).isInstanceOf(Buffer.class)) {
```
ejona86: Whoa. This reads very confusingly. Created https://github.com/netty/netty-incubator-buffer-api/issues/46 to rename `isInstanceOf()`

src/test/java/io/netty/buffer/api/examples/bytetomessagedecoder/AlternativeMessageDecoder.java
```diff
@@ -0,0 +1,115 @@
...
+                && (collector.readableBytes() == 0 || input.readerOffset() == 0)
+                && collector.order() == input.order()) {
+            Buffer.extendComposite(collector, input);
+            drainCollector(ctx);
+            return;
+        }
+        if (collector.isOwned()) {
+            collector.ensureWritable(input.readableBytes(), DEFAULT_CHUNK_SIZE, true);
+        } else {
+            try (Buffer prev = collector) {
```
ejona86: The exception and reference counting interactions here are quite hard to pick apart. I'm assuming this is a try-with-resources for the exception handling. Something like this seems more obviously correct:

```java
int requiredCapacity = input.capacity() + collector.readableBytes();
int newSize = Math.max(requiredCapacity, DEFAULT_CHUNK_SIZE);
try (Buffer newBuffer = allocator.allocate(newSize, input.order())) {
  collector.copyInto(collector.readerOffset(), newBuffer, 0, collector.readableBytes());
  newBuffer.readerOffset(collector.readableBytes());
  collector.close();
  collector = newBuffer.acquire();
}
```

src/main/java/io/netty/buffer/api/CompositeBuffer.java
```diff
@@ -649,13 +652,20 @@ public void ensureWritable(int size, boolean allowCompaction) {
...
+                    return;
+                }
             }
         }
 
-        long newSize = capacity() + (long) size;
-        BufferAllocator.checkSize(newSize);
-        int growth = size - writableBytes();
-        Buffer extension = bufs.length == 0? allocator.allocate(growth) : allocator.allocate(growth, order());
+        int growth = Math.max(size - writableBytes(), minimumGrowth);
```
ejona86, line 666: What prevents doing this in a default method and calling `ensureWritable(int, boolean)`?

src/test/java/io/netty/buffer/api/examples/bytetomessagedecoder/ByteToMessageDecoderTest.java
```diff
@@ -0,0 +1,561 @@
...
+
+        private WriteFailingByteBuf(int untilFailure, Buffer buffer) {
+            super(buffer);
+            this.untilFailure = untilFailure;
+        }
+
+        @Override
+        public Buffer order(ByteOrder order) {
+            if (order == LITTLE_ENDIAN && --untilFailure <= 0) {
+                throw error;
```
ejona86: This seems like a weird need for `BufferAdaptor`. I thought we were going with the idea "people generally wouldn't implement their own Buffer." If that's the case, then injecting an exception into this simple setter seems unnecessary, as it simply wouldn't happen. But if we introduce something like BufferAdaptor, that _encourages_ people to create their own Buffer.

ejona86 (commented):
Adding the argument to `bifurcate()` seems to help a lot. And the docs for slice vs bifurcate seem fair.

In no way was this a full review. I glanced at a few things, and then happened to find some other things while Ctrl-F'ing for something else and noticed some things along the way :smile:.

src/test/java/io/netty/buffer/api/examples/bytetomessagedecoder/ByteToMessageDecoderTest.java
chrisvest: The test was doing something like that, and I wanted to port it over as faithfully as possible. But you're right. We shouldn't have something like that, and the test needs to be adapted, which I can do now that I know it passes as written.

src/test/java/io/netty/buffer/api/examples/bytetomessagedecoder/AlternativeMessageDecoder.java
chrisvest: Yeah, that's better. Thanks :)

src/main/java/io/netty/buffer/api/CompositeBuffer.java
chrisvest: This method will have to do this calculation anyway. The correct default for such an overload is a `minimumGrowth` size of 1.

It wasn't obvious to me which overloads were needed; `(int, boolean)`, `(int, int)`… Adding both of those felt like it would inflate the interface more than necessary.

src/test/java/io/netty/buffer/api/examples/bytetomessagedecoder/AlternativeMessageDecoder.java
chrisvest: I will fix this in another PR.

chrisvest:
@ejona86 I've addressed your comments. Please take a look.

src/test/java/io/netty/buffer/api/examples/bytetomessagedecoder/AlternativeMessageDecoder.java
chrisvest: https://github.com/netty/netty-incubator-buffer-api/pull/49

Issue #47, opened 2021-04-24
Why isn't Composite its own type extending Buffer?


ejona86:
I don't see the point of the Composite API being directly on the Buffer API. Why not move it out to an interface or concrete class with only the additional method `extendComposite()`? It could be created/returned from BufferAllocator directly. Using a final class would allow safely managing unwrapping of composites during extendComposite, but there's obviously tradeoffs there.

I'd generally consider it to be a code smell to call `isComposite()` on a buffer the code didn't create, which means there should very rarely be any calls to `isComposite()`. For most usages, returning the clear type from BufferAllocator seems clearer.

If there's doubt as to which approach is best, maybe the decision here will become more obvious as more code is written using the new API.

jameskleeh:
Agreed. Having `Buffer.isComposite` is not a great design imo. If you need to know if its a composite then that should indicate it should be a separate public class.

chrisvest:
The compose method was initially on the BufferAllocator interface, but since there's only one fixed implementation, it didn't make sense to let the allocator implementation decide how buffer composition works. I think we still need to create composite buffers with a static factory method somewhere.

I think at this point, with a compose overload to deal with #45, we would have 4 methods that are composite buffer specific, so maybe collecting them in their own interface makes sense. Should this interface be only those static methods, or should composite buffers implement this interface? If the latter, should the interface extend Buffer? As much as possible I want buffers to behave the same regardless of implementation, but composite buffers have a few important differences to them, and I haven't yet found an argument that nails that composite buffers must be done one way or another.

ejona86:
> The compose method was initially on the BufferAllocator interface, but since there's only one fixed implementation, it didn't make sense to let the allocator implementation decide how buffer composition works. I think we still need to create composite buffers with a static factory method somewhere.

If composite buffer was a final class then the allocator wouldn't get to choose the implementation, only how to construct it. I guess `compose()` is less interesting for this issue than `extendComposite()`. We could leave the current Buffer.compose() method and just have it return a Composite type. I do find it to be weird for Buffer to depend on Allocator and vise versa, sort of needlessly, but I don't think it has much functional impact and we may end up bikeshedding about that detail.

> Should this interface be only those static methods

I think this issue was "why are there these static methods for something that is clearly a type-specific method?" So there should be some non-static methods. Once you deal with the non-static methods, the static ones may fall into place.

[...]

PR #49, opened 2021-04-26
Rename Deref.isInstanceOf to referentIsInstanceOf

### LINKED ISSUE #46: Send.isInstanceOf is a confusing/misleading name
In reviewing #44, I was very surprised when I ended up comprehending what this line did:
```
        } else if (msg instanceof Send && ((Send<?>) msg).isInstanceOf(Buffer.class)) {
            try (Buffer input = ((Send<Buffer>) msg).receive()) {
```

`isInstanceOf` should probably be renamed as it is far too similar to `Class.isInstance()`, _and_ `msg` _isn't_ an instance of Buffer. `referenceInstanceOf` or something.

chrisvest:
Motivation:
Derefs are not necessarily their referents.
This is the case for Send, for instance.

Modification:
The Deref.isInstanceOf method is renamed to referentIsInstanceOf.
And a Send.isSendOf method has been added, that simplifies the check for sends, since it could be the case that one also needs to check if the object in question is also a Send instance.

Result:
Cleaner code that is easier to read, when working with Sends.

This fixes #46

Please take a look, @ejona86.

### DIFF
--- a/src/main/java/io/netty/buffer/api/Send.java
+++ b/src/main/java/io/netty/buffer/api/Send.java
@@ -56,7 +56,7 @@ public interface Send<T extends Rc<T>> extends Deref<T> {
             }
 
             @Override
-            public boolean isInstanceOf(Class<?> cls) {
+            public boolean referentIsInstanceOf(Class<?> cls) {
                 return cls.isAssignableFrom(concreteObjectType);
             }
 
@@ -67,6 +67,19 @@ public interface Send<T extends Rc<T>> extends Deref<T> {
                 }
             }
         };
+    }
+
+    /**
+     * Determine if the given candidate object is an instance of a {@link Send} from which an object of the given type
+     * can be received.
+     *
+     * @param type The type of object we wish to receive.
+     * @param candidate The candidate object that might be a {@link Send} of an object of the given type.
+     * @return {@code true} if the candidate object is a {@link Send} that would deliver an object of the given type,
+     * otherwise {@code false}.
+     */
+    static boolean isSendOf(Class<?> type, Object candidate) {
+        return candidate instanceof Send && ((Send<?>) candidate).referentIsInstanceOf(type);
     }
 
     /**
--- a/src/main/java/io/netty/buffer/api/Deref.java
+++ b/src/main/java/io/netty/buffer/api/Deref.java
@@ -48,5 +48,5 @@ public interface Deref<T extends Rc<T>> extends Supplier<T> {
      * @param cls The type to check.
      * @return {@code true} if the object in this {@code Deref} can be assigned fields or variables of the given type.
      */
-    boolean isInstanceOf(Class<?> cls);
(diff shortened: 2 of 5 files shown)


src/main/java/io/netty/buffer/api/Send.java
```diff
@@ -56,7 +56,7 @@ public T receive() {
             }
 
             @Override
-            public boolean isInstanceOf(Class<?> cls) {
+            public boolean referentIsInstanceOf(Class<?> cls) {
                 return cls.isAssignableFrom(concreteObjectType);
```
ejona86, line 60: This looks reversed. `concreteObjectType.isAssignableFrom(cls)`. I think the ordering is opposite `isInstance()`. (`a.isInstance(b) == true` says b can be assigned to a variable of type a; `a.isAssignableFrom(b)` says a can be assigned to a variable of type b). I noticed because `Send.isSendOf(Object.class, x)` should have always returned true if we are using `instanceof` semantics.

That said, as long as we are using instanceof semantics, the caller should be casting to `Send<? extends X>` instead of `Send<X>`. Although that is more verbose, I'm not sure it actually hurts anything.

Note that changing to class equality `cls.equals(concreteObjectType)` (such that we can cast to `Send<X>`) would be a _very good_ reason not to have a separate composite class type (https://github.com/netty/netty-incubator-buffer-api/issues/47).

chrisvest: The same code is in `TransferSend`, and flipping the order there makes the `BufferSendTest` fail.

According to the `isAssignableFrom` javadoc:

> Determines if the class or interface represented by this Class object is either the same as, or is a superclass or superinterface of, the class or interface represented by the specified Class parameter.

So the ordering is correct as is.

If we used class equality, then we wouldn't be able to ask about interfaces of the object, and `Buffer` is an interface, so I don't think that would fly.

I'll share my thoughts on the composite class type on that issue.

ejona86: I missed that `assertFalse(Send.isSendOf(Object.class, new Object()));` was passing `new Object()` and not bufferSend/bufferRefSend.

I had thought that isAssignableFrom and isInstance behaved in the same order, and reads the docs once and thought it agreed with me. But then I looked _a second time_ and mis-read the second paragraph of the text and re-read it all and then convinced myself of the other direction. /sigh

> If we used class equality, then we wouldn't be able to ask about interfaces of the object, and Buffer is an interface, so I don't think that would fly.

No, you can still do it for Buffer. But the object would need to know the type of generic it has. That would mean it couldn't be a default method on Rc, but otherwise would be fine.

As we write more code, we may want to keep watch of whether we want a method that will do the cast for you after checking the provided Class. That could return a proper-generics `Send<? extends X>` as well as silence unchecked cast warnings.

ejona86 (approved):

PR #54, opened 2021-04-30
Buffers as constants

chrisvest:
Explains how to correctly implement the buffer-as-a-constant use case, and provide APIs that help do this correctly.


chrisvest:
Please take a look, @njhill

src/main/java/io/netty/buffer/api/BufferAllocator.java
```diff
@@ -68,6 +69,38 @@ default Buffer allocate(int size, ByteOrder order) {
...
+     * in unexpected ways, since the same buffer instance is shared and accessed from many places. The buffer could also
+     * be deallocated, making the data inaccessible. Lastly, the read-only state could be changed, allowing the
+     * supposedly constant buffer to change its contents. The supplier-based API solves all of these problems, by
+     * enforcing that each usage get their own distinct buffer instance.
+     *
+     * @param bytes The byte contents of the buffers produced by the returned supplier.
+     * @return A supplier of read-only buffers with the given contents.
+     */
+    default Supplier<Buffer> constBufferSupplier(byte[] bytes) {
+        return () -> allocate(bytes.length).writeBytes(bytes).readOnly(true);
```
normanmaurer: I wonder if we should have a method for direct and one for heap buffers ?

chrisvest: This is currently given by the choice of allocator. The allocate method will be either on or off heap depending on the instance this is called on.

chrisvest:
@normanmaurer @njhill @ejona86 Optimised const buffer implementations have been pushed now.

src/main/java/io/netty/buffer/api/BufferAllocator.java
```diff
@@ -68,6 +69,41 @@ default Buffer allocate(int size, ByteOrder order) {
...
```
ejona86: So `bytes` is retained and read from after the method call returns. That may be surprising.

ejona86 (commented):
Copy-on-write is interesting, but it seems that isn't the normal behavior given to `readOnly(boolean)`. What _is_ the normal expected use-case of `readOnly(boolean)`? When would code call `readOnly(false)`?

Honestly, I think this problem is just evidence that the ownership model is wrong. If you don't own a Buffer, the Buffer is nerfed and is mostly unusable. The general approach in practice I expect will be "never make buffers unowned" and this change further confirms it. I think it needs to be addressed head-on: either change the ownership/refcounting model so buffers don't become useless when backing memory is shared or change the model so there is no need to expose when backing memory is shared such that they are always owned. For example, to avoid exposing shared backing memory, this copy-on-write approach could be general-purpose.

src/main/java/io/netty/buffer/api/BufferAllocator.java
chrisvest: Yeah, that should be a defensive copy.

chrisvest:
Support for `readOnly(false)` was added to that pooling allocators could fully reset the state of their pooled buffers. However, the pooling allocator doesn't actually pool Buffer objects anymore. Instead, it pools some opaque Buffer implementation-specific objects that can be turned into buffers. This means the read-only flag could be made a one-way street. This would make the read-only protection stronger, though it wouldn't prevent the buffer from being split, or from getting conflicting updates to its offsets when improperly shared. With that in mind I think we will still want this Supplier-based API.

Making read-only irreversible can still be done in a number of ways:

* We can change the state of the buffer in-place. This is the most similar to the current behaviour, and would be my preferred way. We can only have writable access to the memory if we keep around a slice that we made earlier.
* We can return a new buffer instance, like a slice, that is read-only. This is like how ByteBuf currently works, except ByteBuf doesn't increase its reference count in this case. This would in the new API leave us with two buffers that need to be closed, which could be awkward or error prone.
* We can make the read-only state a wrapper, that doesn't increase the reference count. This is exactly like how ByteBuf does it, except the ReadOnlyByteBuf has an unwrap method, which I don't think we would want to replicate. I'm not a fan of increasing indirections in the implementation, and it would also be an odd exception to the reference counting in the new API.
* We can remove the ability to change the read-only state entirely, and have the const buffer supplier be the only way to make them. I wonder if this is too restrictive, though I can't think of a use case for read-only buffers outside of the one addressed by the const buffer supplier already.

With irreversible read-only, we would no longer need to make defensive data copies when making slices of buffers.

I'm guessing that the "evidence that the ownership model is wrong" that you see, is because of the work required to make the supplier return owned buffers? I think that's orthogonal because we'd still have to that copying in order to prevent the buffers from interfering with each other, even if ownership wasn't something we tracked.

That doesn't mean that the ownership model can't be wrong, though, so I will open another issue for that.

chrisvest:
Merging. We can address more comments in future PRs.

PR #60, opened 2021-05-06
Make the build use less space on the CI host

chrisvest:
High space usage could cause the docker layer cache to fail while packaging the layers.


chrisvest:
Hmm… thinking about it, the issue with the layer cache stashing probably occurs because it tries to also stash layers that were downloaded at the beginning, but never actually used in the build.

PR #61, opened 2021-05-07
Add splitComponentsFloor and splitComponentsCeil

### LINKED ISSUE #48: Composite buffer API
We also need a way to release a section of a composite buffer. This need and the current shortcomings of the current file upload API are described here netty/netty#8655. See https://github.com/micronaut-projects/micronaut-core/blob/2.4.x/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpDataReference.java#L221-L243

Basically the composite buffer API currently has methods to retrieve and release individual buffers within the composite. We need this functionality to support streaming file uploads.

chrisvest:
These methods make it possible to accurately split composite buffers at component boundaries, either by rounding the offset down or up to the nearest component boundary, respectively.

Composite buffers already support the split method, but it is hard for client code to predict precisely where component boundaries are placed inside composite buffers.
When split is used with an offset that does not land exactly on a component boundary, then the internal component that the offset lands on will also be split.
This may make it harder to precisely reason about memory life cycles and reuse.

This fixes #48

### DIFF
--- a/src/main/java/io/netty/buffer/api/Buffer.java
+++ b/src/main/java/io/netty/buffer/api/Buffer.java
@@ -607,7 +607,8 @@ public interface Buffer extends Rc<Buffer>, BufferAccessors {
      * See the <a href="#slice-split">Slice vs. Split</a> section for details on the difference between slice
      * and split.
      *
-     * @return A new buffer with independent and exclusive ownership over the read and readable bytes from this buffer.
+     * @return A new buffer with independent and exclusive ownership over the bytes from the beginning to the given
+     * offset of this buffer.
      */
     Buffer split(int splitOffset);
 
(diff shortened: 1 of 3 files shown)


chrisvest:
@jameskleeh What do you think?

jameskleeh:
@chrisvest If a buffer is split in this way, how can the original component be removed from the original buffer?

chrisvest:
@jameskleeh Can you explain your question in more detail?

jameskleeh:
@chrisvest Sure - so the use case is an upload is made up of a composite buffer and the functionality we support is to read a chunk of the file at a time (a component in the buffer is a chunk in this case). So we associate a component in the buffer with an object the user has access to. When the user calls `chunk.getBytes()` for example, the buffer is converted to a byte[] and the component is removed from the upload. This immediately frees the memory so it allows for in memory upload of files of unlimited size. If the component is not removed from the buffer then the composite buffer can not be larger than I think 2GB due to the max size of an integer used within the implementation to store the offset and length.

chrisvest:
@jameskleeh The split methods do remove components from the buffer you called split on. The removed components are contained in the returned buffer. You can then convert that to a byte array, or get `ByteBuffer` views with `forEachReadable`.

chrisvest:
Merging this because I think the rest is mostly about how HTTP upload APIs are designed.

Issue #56, opened 2021-05-04
Ownership model discussion


chrisvest:
Copying @ejona86 from https://github.com/netty/netty-incubator-buffer-api/pull/54#pullrequestreview-650525306

> Honestly, I think this problem is just evidence that the ownership model is wrong. If you don't own a Buffer, the Buffer is nerfed and is mostly unusable. The general approach in practice I expect will be "never make buffers unowned" and this change further confirms it. I think it needs to be addressed head-on: either change the ownership/refcounting model so buffers don't become useless when backing memory is shared or change the model so there is no need to expose when backing memory is shared such that they are always owned. For example, to avoid exposing shared backing memory, this copy-on-write approach could be general-purpose.

Let's take a look. Suppose we want buffers to have ownership at all times, what would that look like? There are only two operations that can cause a buffer to loose ownership: acquire and slice. We will have to get rid of these in order to let buffers have ownership at all times.

We could probably live without acquire without too much trouble, but slice is a little harder. The slice methods let you, perhaps temporarily, focus on a sub-section of a buffer, for reading (IO into buffer), writing (IO out of buffer), or processing. I see the following alternatives to slicing:

* Rely on offsets. Buffers already have reader and writer offsets, and we could also pass explicit offset and length arguments to any method that operates on a buffer. This has precedent with IO methods, so those use cases are perhaps not a big worry. Processing comes in two varieties, as I see it. The first is to break a larger buffer into smaller parts, such as frames, packets, objects, chunks. This could also be solved with split. The second involve doing some work on a small section of the buffer, as part of some larger processing. There are not many examples of this in Netty. This could be decoding a length prefixed string, making a buffer of zeros of a particular length, searching for some header. These could be solved by passing explicit length arguments. That's something that would then need to be taken into account every such an API is created.
* Always copy. Instead of slices sharing the underlying memory with their parent, we could produce buffers that contain copies of the relevant sections of memory. This might work for writing and more processing, but it would not work for reading. IO would have to rely on offsets in this case. A drawback of this approach is the overhead involved in all the memory copies.
[...]

PR #62, opened 2021-05-07
Hide Rc.countBorrows

chrisvest:
The state that people really care about is whether or not an Rc has ownership.
Exposing the reference count will probably just confuse people.
The reference count is still exposed on RcSupport because it may be (and is, in the case of ByteBufAdaptor) needed to support implementation details.

### DIFF
--- a/src/main/java/io/netty/buffer/api/BufferHolder.java
+++ b/src/main/java/io/netty/buffer/api/BufferHolder.java
@@ -78,11 +78,6 @@ public abstract class BufferHolder<T extends BufferHolder<T>> implements Rc<T> {
         return buf.isOwned();
     }
 
-    @Override
-    public int countBorrows() {
-        return buf.countBorrows();
-    }
-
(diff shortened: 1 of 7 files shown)


chrisvest:
@NiteshKant this what you had in mind?

normanmaurer:
I like this!

normanmaurer (approved):

NiteshKant (approved):
@chrisvest this is good. 

I am still wrapping my head around things here so I am not sure if `send()` and `acquire()` are redundant too. In my mind, if we could get rid of ref count as a concept completely in public APIs and documentation, it will make things simpler.

chrisvest:
Then we'd have to get rid of the slice methods as well. That might be a tough sell.

ejona86 (approved):

ejona86:
@NiteshKant getting rid of explicit ref count is my ideal as well. But I think that's a conversation for #56.

PR #70, opened 2021-05-26
Remove the concept of reference counting

### LINKED ISSUE #56: Ownership model discussion

### LINKED ISSUE #69: Rename `Rc` to `ReferenceCounted`
I did minimal searching before opening this issue, so my apologies if this has been discussed previously.

Currently this class (`Rc`) exists in Netty 4.1 as [`ReferenceCounted`](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/ReferenceCounted.java), but lost the descriptive name to something that I would call confusing: [`Rc`](https://github.com/netty/netty-incubator-buffer-api/blob/main/buffer-api/src/main/java/io/netty/buffer/api/Rc.java).

I would propose to rename this interface back to `ReferenceCounted`.

chrisvest:
This will simplify the API a great deal, hopefully.

- [x] Remove `acquire`
- [x] Get tests passing again
- [x] Remove `slice`
- [x] Add `copy` methods (allocates new buffer and copy data to it)
- [x] Get tests passing again
- [x] Remove `Rc` interface


hyperxpro:
This is really required. TBH, Reference counting is just painful to deal with.

RATIONALE.adoc
```diff
@@ -71,32 +71,15 @@ Hopefully you’ll be able to see these principles reflected in the new API.
...
-Every allocation and acquire call on a buffer (any `Rc` object, really) must be paired with a `close()`, and every `receive()` call on a `Send` must also be paired with a `close()`.
-Every method that increments the reference count will document this fact.
-When the reference count is not incremented, then there is obviously no need to decrement it either.
-Buffers can thus be passed through a pipeline without having to do much reference counting work.
+Every allocation and acquire call on a buffer (any `Resource` object, really) must be paired with a `close()`, and every `receive()` call on a `Send` must also be paired with a `close()`.
+
+While referene counting is a useful thing for tracking resource life-cycles internally, it is not itself exposed in the public API.
+Instead, the public API effectively has a boolean open/closed state.
+This simplifies the API a great deal; buffers are created, and in the end they are closed.
+The code in between needs to be arranged such that it just avoids ever holding on to buffers that might be closed.
```
normanmaurer, line 82: I like this... this also removes the pattern where sometimes people do something like:

```
while (buf.refCnt() > 0) {
    buf.release();
}
```
I never saw a use-case where this not just did hide a bug somewhere else.

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -100,35 +74,20 @@
...
- *         They can do this because the regions are guaranteed to not overlap; data changes through one buffer will not
- *         be visible through the other.
- *     </li>
- * </ul>
- *
- * These differences mean that slicing is mostly suitable for when you temporarily want to share a focused area of a
- * buffer.
- * Examples of this include doing IO, or decoding a bounded part of a larger message.
- * On the other hand, split is suitable for when you want to hand over a region of a buffer to some other,
+ * The {@link #split()} method break a buffer into two.
```
normanmaurer, line 79: nit: break -> breaks

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -100,35 +74,20 @@
...
+ * The two buffers will share the underlying memory, but their regions will not overlap, ensuring that the memory is
+ * safely shared between the two.
+ * <p>
+ * Splitting a buffer is useful for when you want to hand over a region of a buffer to some other,
  * perhaps unknown, piece of code, and relinquish your ownership of that buffer region in the process.
  * Examples include aggregating messages into an accumulator buffer, and sending messages down the pipeline for
  * further processing, as split buffer regions, once their data has been received in its entirety.
  *
+ * If you instead wish to temporarily share a region of a buffer, you will have to pass offset and length along with the
+ * buffer, or you will have to make a copy of the region in a new buffer.
```
normanmaurer, line 89: nit: I think I would remove the "in a new buffer".

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -412,16 +370,14 @@ default ByteCursor openReverseCursor() {
...
      * If this buffer already has the necessary space, then this method returns immediately.
      * If this buffer does not already have the necessary space, then it will be expanded using the
      * {@link BufferAllocator} the buffer was created with.
      * This method is the same as calling {@link #ensureWritable(int, int, boolean)} where {@code allowCompaction} is
      * {@code false}.
      *
      * @param size The requested number of bytes of space that should be available for writing.
-     * @throws IllegalStateException if this buffer is not in an {@linkplain #isOwned() owned} state,
-     * or is {@linkplain #readOnly() read-only}.
+     * @throws IllegalStateException if this buffer is not in a bad state, or is {@linkplain #readOnly() read-only}.
```
normanmaurer, line 380: I think this should be `ff this buffer is in a bad state` (remove the not)

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -463,59 +418,46 @@ default void ensureWritable(int size) {
      * @param allowCompaction {@code true} if the method is allowed to modify the
      *                                   {@linkplain #readerOffset() reader offset} and
      *                                   {@linkplain #writerOffset() writer offset}, otherwise {@code false}.
-     * @throws IllegalStateException if this buffer is not in an {@linkplain #isOwned() owned} state,
-     * or is {@linkplain #readOnly() read-only}.
+     * @throws IllegalStateException if this buffer is not in a bad state, or is {@linkplain #readOnly() read-only}.
```
normanmaurer, line 421: same comment as above...

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -615,8 +553,7 @@ default Buffer split() {
     /**
      * Discards the read bytes, and moves the buffer contents to the beginning of the buffer.
      *
-     * @throws IllegalStateException if this buffer is not in an {@linkplain #isOwned() owned} state,
-     * or is {@linkplain #readOnly() read-only}.
+     * @throws IllegalStateException if this buffer is not in a bad state, or is {@linkplain #readOnly() read-only}.
```
normanmaurer, line 556: same comment as above... and same comment everywhere :D

buffer-api/src/main/java/io/netty/buffer/api/BufferHolder.java
```diff
@@ -125,27 +96,9 @@ protected final void replaceBuf(Buffer newBuf) {
...
-     * The buffer assignment is performed using a volatile store.
-     *
-     * @param newBuf The new {@link Buffer} instance that is replacing the currently held buffer.
-     */
-    protected final void replaceBufVolatile(Buffer newBuf) {
-        var prev = (Buffer) BUF.getAndSet(this, newBuf.acquire());
-        prev.close();
+    protected final void replaceBuffer(Send<Buffer> send) {
+        buf.close();
+        buf = send.receive();
```
normanmaurer, line 101: Could the `receive()` throw ? It might be safer to only call `close` once we know we can receive a new buffer otherwise we might end up in a strange state

buffer-api/src/main/java/io/netty/buffer/api/Resource.java
```diff
@@ -0,0 +1,41 @@
+package io.netty.buffer.api;
```
normanmaurer, line 1: nit: license header

buffer-api/src/main/java/io/netty/buffer/api/Resource.java
```diff
@@ -0,0 +1,41 @@
...
+     * within a try-with-resources statement.
+     */
+    Send<T> send();
+
+    /**
+     * Close the resource, making it inaccessible.
+     * <p>
+     * Note, this method is not thread-safe unless otherwise specific.
+     *
+     * @throws IllegalStateException If this {@code Resource} has already been closed.
```
normanmaurer, line 28: I know this is ok from an API contract but I wonder if we should better follow what is stated in the javadocs:

```
Note that unlike the close method of Closeable, this close method is not required to be idempotent. In other words, calling this close method more than once may have some visible side effect, unlike Closeable.close which is required to have no effect if called more than once. However, implementers of this interface are strongly encouraged to make their close methods idempotent.

```

buffer-api/src/main/java/io/netty/buffer/api/adaptor/ByteBufAdaptor.java
```diff
@@ -1416,11 +1419,8 @@ public ByteBuf slice(int index, int length) {
     @Override
     public ByteBuf retainedSlice(int index, int length) {
         checkAccess();
-        try {
-            return wrap(buffer.slice(index, length));
-        } catch (IllegalStateException e) {
-            throw new IllegalReferenceCountException(e);
-        }
+        retain();
```
normanmaurer, line 1422: is there any chance that ` new Slice` might throw ? If so would it be better to call retain() after it ?

buffer-api/src/main/java/io/netty/buffer/api/internal/ResourceSupport.java
```diff
@@ -92,28 +108,35 @@ public final void close() {
...
                 "Cannot send() a reference counted object with " + countBorrows() + " borrows: " + this + '.');
     }
 
-    @Override
-    public boolean isOwned() {
+    static boolean isOwned(ResourceSupport<?, ?> obj) {
+        return obj.isOwned();
+    }
+
+    protected boolean isOwned() {
```
normanmaurer, line 124: javadocs for protected methods ?

chrisvest:
@normanmaurer I'll address these comments in #72

buffer-api/src/main/java/io/netty/buffer/api/Resource.java
chrisvest: I think calling `close` more than once is indicative of a bug, where there's uncertainty about the buffers life time. The `release` method on `ByteBuf` also throws if you call it on a buffer that's already been deallocated.

buffer-api/src/main/java/io/netty/buffer/api/internal/ResourceSupport.java
chrisvest: This completely internal, but I can add some

PR #71, opened 2021-05-28
chore: fix a few typos

kashike:
A few mistakes I found while looking through the code this morning.


buffer-api/src/main/java/io/netty/buffer/api/bytebuffer/NioBuffer.java
```diff
@@ -98,19 +98,19 @@ public void attach(NioBuffer buf) {
 
         @Override
         public String toString() {
-            return "MemSegDrop(" + delegate + ')';
+            return "MakeInaccessibleOnDrop(" + delegate + ')';
```
kashike, line 101: I changed this in `UnsafeBuffer` as well - they were both very different. Is this good, or should we use some other naming?

chrisvest: These toStrings are only for debugging, so as long as it's clear to a human with a debugger it's fine. 👍

chrisvest (approved):

PR #72, opened 2021-05-28
Implement BufferIntegratable for Nio- and UnsafeBuffer

chrisvest:
This fixes #64

PR #73, opened 2021-05-28
Move to custom exception types

kashike:
Opening this to get some initial feedback before continuing.

Rather than using a single type (`IllegalStateException`) when notifying consumers about exceptions, use custom ones - this makes catching and handling exceptions easier.

### DIFF
--- a/buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
+++ b/buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
@@ -149,7 +149,8 @@ public interface Buffer extends Resource<Buffer>, BufferAccessors {
      * @return This Buffer.
      * @throws IndexOutOfBoundsException if the specified {@code offset} is less than the current
      * {@link #readerOffset()} or greater than {@link #capacity()}.
-     * @throws IllegalStateException if this buffer is {@linkplain #readOnly() read-only}.
+     * @throws BufferClosedException if this buffer is closed.
+     * @throws BufferReadOnlyException if this buffer is {@linkplain #readOnly() read-only}.
      */
     Buffer writerOffset(int offset);
 
@@ -174,7 +175,7 @@ public interface Buffer extends Resource<Buffer>, BufferAccessors {
      *
      * @param value The byte value to write at every position in the buffer.
      * @return This Buffer.
-     * @throws IllegalStateException if this buffer is {@linkplain #readOnly() read-only}.
+     * @throws BufferReadOnlyException if this buffer is {@linkplain #readOnly() read-only}.
      */
     Buffer fill(byte value);
 
@@ -377,7 +378,9 @@ public interface Buffer extends Resource<Buffer>, BufferAccessors {
      * {@code false}.
      *
      * @param size The requested number of bytes of space that should be available for writing.
-     * @throws IllegalStateException if this buffer is in a bad state, or is {@linkplain #readOnly() read-only}.
+     * @throws IllegalStateException if this buffer is in a bad state.
+     * @throws BufferClosedException if this buffer is closed.
+     * @throws BufferReadOnlyException if this buffer is {@linkplain #readOnly() read-only}.
      */
     default void ensureWritable(int size) {
         ensureWritable(size, 1, true);
@@ -418,7 +421,9 @@ public interface Buffer extends Resource<Buffer>, BufferAccessors {
      * @param allowCompaction {@code true} if the method is allowed to modify the
      *                                   {@linkplain #readerOffset() reader offset} and
      *                                   {@linkplain #writerOffset() writer offset}, otherwise {@code false}.
-     * @throws IllegalStateException if this buffer is in a bad state, or is {@linkplain #readOnly() read-only}.
+     * @throws BufferReadOnlyException if this buffer is {@linkplain #readOnly() read-only}.
+     * @throws IllegalArgumentException if {@code size} or {@code minimumGrowth} are negative.
+     * @throws IllegalStateException if this buffer is in a bad state.
      */
     void ensureWritable(int size, int minimumGrowth, boolean allowCompaction);
 
@@ -553,7 +558,8 @@ public interface Buffer extends Resource<Buffer>, BufferAccessors {
     /**
      * Discards the read bytes, and moves the buffer contents to the beginning of the buffer.
      *
-     * @throws IllegalStateException if this buffer is in a bad state, or is {@linkplain #readOnly() read-only}.
+     * @throws BufferReadOnlyException if this buffer is {@linkplain #readOnly() read-only}.
+     * @throws IllegalStateException if this buffer is in a bad state.
      */
     void compact();
 
--- a/buffer-api/src/main/java/io/netty/buffer/api/BufferClosedException.java
+++ b/buffer-api/src/main/java/io/netty/buffer/api/BufferClosedException.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer.api;
+
+/**
+ * An exception thrown when an operation is attempted on a {@link Buffer} when it has been closed.
+ */
+public final class BufferClosedException extends UnsupportedOperationException {
+    private static final long serialVersionUID = 85913332711192868L;
+
+    public BufferClosedException(final String message) {
+        super(message);
+    }
+}
--- a/buffer-api/src/main/java/io/netty/buffer/api/BufferReadOnlyException.java
+++ b/buffer-api/src/main/java/io/netty/buffer/api/BufferReadOnlyException.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer.api;
+
+/**
+ * An exception thrown when an operation is attempted on a {@linkplain Buffer#readOnly() read-only} {@link Buffer}.
+ */
+public final class BufferReadOnlyException extends UnsupportedOperationException {
+    private static final long serialVersionUID = 4855825594125231593L;
+
+    public BufferReadOnlyException(final String message) {
+        super(message);
+    }
+}
--- a/buffer-api/src/main/java/io/netty/buffer/api/adaptor/ByteBufAdaptor.java
+++ b/buffer-api/src/main/java/io/netty/buffer/api/adaptor/ByteBufAdaptor.java
@@ -23,6 +23,7 @@ import io.netty.buffer.DuplicatedByteBuf;
 import io.netty.buffer.SlicedByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.buffer.api.BufferAllocator;
+import io.netty.buffer.api.BufferReadOnlyException;
 import io.netty.buffer.api.internal.Statics;
 import io.netty.buffer.api.Buffer;
 import io.netty.buffer.api.internal.ResourceSupport;
@@ -253,7 +254,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public byte getByte(int index) {
         try {
             return buffer.getByte(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
@@ -262,7 +263,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public short getUnsignedByte(int index) {
         try {
             return (short) buffer.getUnsignedByte(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
@@ -271,7 +272,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public short getShort(int index) {
         try {
             return buffer.getShort(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
@@ -281,7 +282,7 @@ public final class ByteBufAdaptor extends ByteBuf {
         ByteOrder originalOrder = buffer.order();
         try {
             return buffer.order(ByteOrder.LITTLE_ENDIAN).getShort(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         } finally {
             buffer.order(originalOrder);
@@ -292,7 +293,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public int getUnsignedShort(int index) {
         try {
             return buffer.getUnsignedShort(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
@@ -302,7 +303,7 @@ public final class ByteBufAdaptor extends ByteBuf {
         ByteOrder originalOrder = buffer.order();
         try {
             return buffer.order(ByteOrder.LITTLE_ENDIAN).getUnsignedShort(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         } finally {
             buffer.order(originalOrder);
@@ -313,7 +314,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public int getMedium(int index) {
         try {
             return buffer.getMedium(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
@@ -323,7 +324,7 @@ public final class ByteBufAdaptor extends ByteBuf {
         ByteOrder originalOrder = buffer.order();
         try {
             return buffer.order(ByteOrder.LITTLE_ENDIAN).getMedium(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         } finally {
             buffer.order(originalOrder);
@@ -334,7 +335,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public int getUnsignedMedium(int index) {
         try {
             return buffer.getUnsignedMedium(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
@@ -344,7 +345,7 @@ public final class ByteBufAdaptor extends ByteBuf {
         ByteOrder originalOrder = buffer.order();
         try {
             return buffer.order(ByteOrder.LITTLE_ENDIAN).getUnsignedMedium(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         } finally {
             buffer.order(originalOrder);
@@ -355,7 +356,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public int getInt(int index) {
         try {
             return buffer.getInt(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
@@ -365,7 +366,7 @@ public final class ByteBufAdaptor extends ByteBuf {
         ByteOrder originalOrder = buffer.order();
         try {
             return buffer.order(ByteOrder.LITTLE_ENDIAN).getInt(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         } finally {
             buffer.order(originalOrder);
@@ -376,7 +377,7 @@ public final class ByteBufAdaptor extends ByteBuf {
     public long getUnsignedInt(int index) {
         try {
             return buffer.getUnsignedInt(index);
-        } catch (IllegalStateException e) {
+        } catch (BufferReadOnlyException e) {
             throw new IllegalReferenceCountException(e);
         }
     }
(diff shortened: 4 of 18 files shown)


kashike:
@chrisvest When you have a minute, please take a look and see what you think before I continue. Started this before the comments in #70.

chrisvest:
@kashike I like the idea. I think the exceptions should be defined in each their own class, though.

buffer-api/src/main/java/io/netty/buffer/api/internal/ResourceSupport.java
```diff
@@ -52,7 +53,7 @@ protected ResourceSupport(Drop<T> drop) {
      */
     protected final I acquire() {
         if (acquires < 0) {
-            throw attachTrace(new IllegalStateException("This resource is closed: " + this + '.'));
+            throw attachTrace(new BufferClosedException("This resource is closed: " + this + '.'));
```
chrisvest: `ResourceSupport` should not be buffer specific. We can delegate the exception creation to a protected abstract method that subclasses override. Then we can also make the exception message be "This _buffer_ is closed…"

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -377,7 +378,9 @@ default ByteCursor openReverseCursor() {
      * {@code false}.
      *
      * @param size The requested number of bytes of space that should be available for writing.
-     * @throws IllegalStateException if this buffer is not in a bad state, or is {@linkplain #readOnly() read-only}.
+     * @throws IllegalStateException if this buffer is not in a bad state.
```
chrisvest: Remove "not" – makes the rebase easier when #72 and this eventually conflicts.

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -418,7 +421,9 @@ default void ensureWritable(int size) {
      * @param allowCompaction {@code true} if the method is allowed to modify the
      *                                   {@linkplain #readerOffset() reader offset} and
      *                                   {@linkplain #writerOffset() writer offset}, otherwise {@code false}.
-     * @throws IllegalStateException if this buffer is not in a bad state, or is {@linkplain #readOnly() read-only}.
+     * @throws BufferReadOnlyException if this buffer is {@linkplain #readOnly() read-only}.
+     * @throws IllegalArgumentException if {@code size} or {@code minimumGrowth} are negative.
+     * @throws IllegalStateException if this buffer is not in a bad state.
```
chrisvest: Remove "not" – makes the rebase easier when #72 and this eventually conflicts.

buffer-api/src/main/java/io/netty/buffer/api/Buffer.java
```diff
@@ -553,7 +558,8 @@ default Buffer split() {
     /**
      * Discards the read bytes, and moves the buffer contents to the beginning of the buffer.
      *
-     * @throws IllegalStateException if this buffer is not in a bad state, or is {@linkplain #readOnly() read-only}.
+     * @throws BufferReadOnlyException if this buffer is {@linkplain #readOnly() read-only}.
+     * @throws IllegalStateException if this buffer is not in a bad state.
```
chrisvest: Remove "not" – makes the rebase easier when #72 and this eventually conflicts.

chrisvest (changes requested):

chrisvest:
Also, I've no idea why test report uploading keeps failing on your PRs but works on mine 🤷

normanmaurer:
@chrisvest I wonder if the uploading this is some sort of a permission problem. I had the same problem in netty at some point and so switched to the extra workflow for it:

https://github.com/netty/netty/blob/4.1/.github/workflows/ci-pr-reports.yml

chrisvest:
@kashike I merged a separate workflow for uploading test results, so if you rebase _maybe_ that part will work now.

buffer-api/src/main/java/io/netty/buffer/api/BufferClosedException.java
```diff
@@ -0,0 +1,27 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer.api;
+
+/**
+ * An exception thrown when an operation is attempted on a {@link Buffer} when it has been closed.
+ */
+public class BufferClosedException extends IllegalStateException {
```
normanmaurer: I think I would mark all our exceptions as final for now.

buffer-api/src/main/java/io/netty/buffer/api/BufferReadOnlyException.java
```diff
@@ -0,0 +1,27 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer.api;
+
+/**
+ * An exception thrown when an operation is attempted on a {@linkplain Buffer#readOnly() read-only} {@link Buffer}.
+ */
+public class BufferReadOnlyException extends IllegalStateException {
```
normanmaurer: Why we don't just use what the JDK provides here ? 

https://docs.oracle.com/javase/7/docs/api/java/nio/ReadOnlyBufferException.html

At least I would extend what the JDK provides as `UnsupportedOperationException` sounds also more correct.

normanmaurer (changes requested):

buffer-api/src/main/java/io/netty/buffer/api/BufferReadOnlyException.java
kashike: The JDK class does not allow specifying a message.

kashike: I have changed the parent of ours to `UnsupportedOperationException`.

kashike:
I've addressed the comments left earlier.

kashike:
I have had to do some... ugly things to the tests. Due to the exceptions now having different parents, it becomes a mess. Perhaps we should introduce a root exception type (`BufferException extends RuntimeException`) for our exceptions to extend?

We could then also introduce more exception types for what we throw.

buffer-api/src/main/java/io/netty/buffer/api/internal/Statics.java
```diff
@@ -165,12 +167,12 @@ private static void bbputFallback(ByteBuffer dest, int destPos, ByteBuffer src,
...
-    static IllegalStateException bufferIsClosed() {
-        return new IllegalStateException("This buffer is closed.");
+    static BufferClosedException bufferIsClosed(Buffer buffer) {
+        return new BufferClosedException("This buffer is closed.");
     }
 
-    static IllegalStateException bufferIsReadOnly() {
-        return new IllegalStateException("This buffer is read-only.");
+    static BufferReadOnlyException bufferIsReadOnly(Buffer buffer) {
+        return new BufferReadOnlyException("This buffer is read-only.");
```
chrisvest: Why do these two take `Buffer` as arguments now?

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferComponentIterationTest.java
```diff
@@ -357,7 +359,7 @@ public void forEachWritableOnClosedBufferMustThrow(Fixture fixture) {
         try (BufferAllocator allocator = fixture.createAllocator()) {
             Buffer buf = allocator.allocate(8);
             buf.close();
-            assertThrows(IllegalStateException.class, () -> buf.forEachWritable(0, (index, component) -> true));
+            assertThrows(UnsupportedOperationException.class, () -> buf.forEachWritable(0, (index, component) -> true));
```
chrisvest: Why is this not `BufferClosedException`?

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferCompositionTest.java
```diff
@@ -391,7 +392,8 @@ public void composingReadOnlyBuffersMustCreateReadOnlyCompositeBuffer() {
              Buffer b = allocator.allocate(4).makeReadOnly();
              Buffer composite = CompositeBuffer.compose(allocator, a.send(), b.send())) {
             assertTrue(composite.readOnly());
-            verifyWriteInaccessible(composite);
+            verifyWriteInaccessible(composite, BufferReadOnlyException.class, BufferReadOnlyException.class
+            );
```
chrisvest: Stray line break

buffer-api/src/main/java/io/netty/buffer/api/internal/ResourceSupport.java
```diff
@@ -15,6 +15,7 @@
  */
 package io.netty.buffer.api.internal;
 
+import io.netty.buffer.api.BufferClosedException;
```
chrisvest: This import can be removed now.

buffer-api/src/main/java/io/netty/buffer/api/adaptor/ByteBufAdaptor.java
```diff
@@ -253,7 +253,7 @@ public boolean getBoolean(int index) {
     public byte getByte(int index) {
         try {
             return buffer.getByte(index);
-        } catch (IllegalStateException e) {
+        } catch (IllegalStateException | UnsupportedOperationException e) {
```
chrisvest: I think all of these should just catch `BufferClosedException` and let any `BufferReadOnlyException` bubble through. Buffers being read-only is not a reference count issue. The illegal state probably is, though, so let's keep that.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
```diff
@@ -121,7 +122,8 @@ public void compactOnReadOnlyBufferMustThrow(Fixture fixture) {
         try (BufferAllocator allocator = fixture.createAllocator();
              Buffer buf = allocator.allocate(8)) {
             buf.makeReadOnly();
-            assertThrows(IllegalStateException.class, () -> buf.compact());
+            final RuntimeException c1e = assertThrows(RuntimeException.class, () -> buf.compact());
+            assertTrue(c1e instanceof IllegalStateException || c1e instanceof UnsupportedOperationException);
```
chrisvest: We should assert that this specifically throws `BufferReadOnlyException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
```diff
@@ -215,19 +217,22 @@ public void compactOnConstBufferMustNotImpactSiblings(Fixture fixture) {
                  Buffer c = a.copy()) {
                 assertEquals(1, a.readByte());
                 assertEquals(2, a.readByte());
-                assertThrows(IllegalStateException.class, () -> a.compact()); // Can't compact read-only buffer.
+                final RuntimeException c1e = assertThrows(RuntimeException.class, () -> a.compact()); // Can't compact read-only buffer.
+                assertTrue(c1e instanceof IllegalStateException || c1e instanceof UnsupportedOperationException);
```
chrisvest: These should also assert on `BufferReadOnlyException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
```diff
@@ -215,19 +217,22 @@ public void compactOnConstBufferMustNotImpactSiblings(Fixture fixture) {
...
                 assertEquals(3, a.readByte());
                 assertEquals(4, a.readByte());
 
                 assertEquals(1, b.readByte());
                 assertEquals(2, b.readByte());
-                assertThrows(IllegalStateException.class, () -> b.compact()); // Can't compact read-only buffer.
+                final RuntimeException c2e = assertThrows(RuntimeException.class, () -> b.compact()); // Can't compact read-only buffer.
+                assertTrue(c2e instanceof IllegalStateException || c2e instanceof UnsupportedOperationException);
```
chrisvest: 📌

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
```diff
@@ -215,19 +217,22 @@ public void compactOnConstBufferMustNotImpactSiblings(Fixture fixture) {
...
                 assertEquals(3, b.readByte());
                 assertEquals(4, b.readByte());
 
                 assertEquals(1, c.readByte());
                 assertEquals(2, c.readByte());
-                assertThrows(IllegalStateException.class, () -> c.compact()); // Can't compact read-only buffer.
+                final RuntimeException c3e = assertThrows(RuntimeException.class, () -> c.compact()); // Can't compact read-only buffer.
+                assertTrue(c3e instanceof IllegalStateException || c3e instanceof UnsupportedOperationException);
```
chrisvest: 📌

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReferenceCountingTest.java
```diff
@@ -77,7 +77,7 @@ void acquireOnClosedBufferMustThrow(Fixture fixture) {
         try (BufferAllocator allocator = fixture.createAllocator()) {
             var buf = allocator.allocate(8);
             buf.close();
-            assertThrows(IllegalStateException.class, () -> acquire((ResourceSupport<?, ?>) buf));
+            assertThrows(UnsupportedOperationException.class, () -> acquire((ResourceSupport<?, ?>) buf));
```
chrisvest: This should assert on `BufferClosedException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferTestSupport.java
```diff
@@ -323,88 +324,91 @@ static void stopExecutor() throws IOException {
     public static void verifyInaccessible(Buffer buf) {
         verifyReadInaccessible(buf);
 
-        verifyWriteInaccessible(buf);
+        verifyWriteInaccessible(buf, BufferClosedException.class, IllegalStateException.class);
```
chrisvest: This is shining light on a bug in `ensureWritable`. It should throw `BufferClosedException` in these cases, and we can then collapse this to one parameter.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
```diff
@@ -121,7 +122,8 @@ public void compactOnReadOnlyBufferMustThrow(Fixture fixture) {
         try (BufferAllocator allocator = fixture.createAllocator();
              Buffer buf = allocator.allocate(8)) {
             buf.makeReadOnly();
-            assertThrows(IllegalStateException.class, () -> buf.compact());
+            final RuntimeException c1e = assertThrows(RuntimeException.class, () -> buf.compact());
```
chrisvest: This needs to assert on `BufferReadOnlyException` and `NioBuffer` needs to comply.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
chrisvest: Same here. `NioBuffer` is wrong for diverging.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
chrisvest: 📌

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
chrisvest: 📌

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferRefTest.java
```diff
@@ -37,7 +37,8 @@ public void closingBufRefMustCloseOwnedBuf() {
             ref.contents().writeInt(42);
             assertThat(ref.contents().readInt()).isEqualTo(42);
             ref.close();
-            assertThrows(IllegalStateException.class, () -> ref.contents().writeInt(32));
+            final RuntimeException w1e = assertThrows(RuntimeException.class, () -> ref.contents().writeInt(32));
+            assertTrue(w1e instanceof IllegalStateException || w1e instanceof UnsupportedOperationException);
```
chrisvest: This should assert on `BufferClosedException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferRefTest.java
```diff
@@ -49,7 +50,7 @@ public void closingBufRefMustCloseOwnedBufFromSend() {
             ref.contents().writeInt(42);
             assertThat(ref.contents().readInt()).isEqualTo(42);
             ref.close();
-            assertThrows(IllegalStateException.class, () -> ref.contents().writeInt(32));
+            assertThrows(UnsupportedOperationException.class, () -> ref.contents().writeInt(32));
```
chrisvest: This should be `BufferClosedException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferRefTest.java
```diff
@@ -72,11 +73,13 @@ public void mustCloseOwnedBufferWhenReplaced() {
                 ref.replace(buf.send()); // Pass replacement directly.
             }
 
-            assertThrows(IllegalStateException.class, () -> orig.get().writeInt(32));
+            final RuntimeException w1e = assertThrows(RuntimeException.class, () -> orig.get().writeInt(32));
+            assertTrue(w1e instanceof IllegalStateException || w1e instanceof UnsupportedOperationException);
```
chrisvest: These two should assert on `BufferClosedException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferRefTest.java
```diff
@@ -72,11 +73,13 @@ public void mustCloseOwnedBufferWhenReplaced() {
...
             ref.contents().writeInt(42);
             assertThat(ref.contents().readInt()).isEqualTo(42);
             ref.close();
-            assertThrows(IllegalStateException.class, () -> ref.contents().writeInt(32));
+            final RuntimeException w2e = assertThrows(RuntimeException.class, () -> ref.contents().writeInt(32));
+            assertTrue(w2e instanceof IllegalStateException || w2e instanceof UnsupportedOperationException);
```
chrisvest: 📌

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferRefTest.java
```diff
@@ -99,11 +102,13 @@ public void mustCloseOwnedBufferWhenReplacedFromSend() {
...
```
chrisvest: Same for these two. Looks like this test got duplicated somehow (the one above this one is the duplicate, which can be removed).

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferRefTest.java
```diff
@@ -113,7 +118,8 @@ public void sendingRefMustSendBuffer() {
              BufferRef refA = new BufferRef(allocator.allocate(8).send())) {
             refA.contents().writeInt(42);
             Send<BufferRef> send = refA.send();
-            assertThrows(IllegalStateException.class, () -> refA.contents().readInt());
+            final RuntimeException r1e = assertThrows(RuntimeException.class, () -> refA.contents().readInt());
+            assertTrue(r1e instanceof IllegalStateException || r1e instanceof UnsupportedOperationException);
```
chrisvest: This should assert on `BufferClosedException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferTestSupport.java
```diff
@@ -323,88 +324,91 @@ static void stopExecutor() throws IOException {
...
 
         try (BufferAllocator allocator = BufferAllocator.heap();
              Buffer target = allocator.allocate(24)) {
-            assertThrows(IllegalStateException.class, () -> buf.copyInto(0, target, 0, 1));
-            assertThrows(IllegalStateException.class, () -> buf.copyInto(0, new byte[1], 0, 1));
-            assertThrows(IllegalStateException.class, () -> buf.copyInto(0, ByteBuffer.allocate(1), 0, 1));
+            assertThrows(UnsupportedOperationException.class, () -> buf.copyInto(0, target, 0, 1));
+            assertThrows(UnsupportedOperationException.class, () -> buf.copyInto(0, new byte[1], 0, 1));
+            assertThrows(UnsupportedOperationException.class, () -> buf.copyInto(0, ByteBuffer.allocate(1), 0, 1));
```
chrisvest: These `UnsupportedOperationException` can be `BufferClosedException`.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferTestSupport.java
```diff
@@ -323,88 +324,91 @@ static void stopExecutor() throws IOException {
...
-        assertThrows(IllegalStateException.class, () -> buf.openCursor(0, 0));
-        assertThrows(IllegalStateException.class, () -> buf.openReverseCursor());
-        assertThrows(IllegalStateException.class, () -> buf.openReverseCursor(0, 0));
+        assertThrows(UnsupportedOperationException.class, () -> acquire((ResourceSupport<?, ?>) buf));
+        final RuntimeException c1e = assertThrows(RuntimeException.class, () -> buf.copy());
+        assertTrue(c1e instanceof UnsupportedOperationException || c1e instanceof IllegalStateException);
+        assertThrows(UnsupportedOperationException.class, () -> buf.openCursor());
+        assertThrows(UnsupportedOperationException.class, () -> buf.openCursor(0, 0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.openReverseCursor());
+        assertThrows(UnsupportedOperationException.class, () -> buf.openReverseCursor(0, 0));
```
chrisvest: All of these should be `BufferClosedException` as well.

The `split` and `send` methods will need some `isAccessible` checks before their `isOwned` checks.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferTestSupport.java
```diff
@@ -323,88 +324,91 @@ static void stopExecutor() throws IOException {
...
+        assertThrows(UnsupportedOperationException.class, () -> buf.getChar(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getShort(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getUnsignedShort(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getMedium(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getUnsignedMedium(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getInt(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getUnsignedInt(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getFloat(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getLong(0));
+        assertThrows(UnsupportedOperationException.class, () -> buf.getDouble(0));
```
chrisvest: All of these also need to assert on `BufferClosedException` – pure read methods cannot throw `BufferReadOnlyException`.

chrisvest (changes requested):
Thanks for doing this work. I had a bunch of comments. Please take a look.

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
kashike: They don't just throw this, which is why it was changed. 
> org.opentest4j.AssertionFailedError: Unexpected exception type thrown ==> expected: <io.netty.util.IllegalReferenceCountException> but was: <io.netty.buffer.api.BufferClosedException>

buffer-tests/src/test/java/io/netty/buffer/api/tests/BufferReadOnlyTest.java
kashike: What would you prefer, moving the read-only check above the owned check?

kashike:
I've pushed some of the requested changes, with around 200 tests failing locally.

chrisvest (approved):
I fixed the remaining issues

kashike:
Cheers @chrisvest - I was not sure what you had in mind.

PR #74, opened 2021-05-28
feat: Re-use a NOOP tracer

kashike:
This class (and the parent) has no state - there is no reason we cannot reuse an instance of `NoOpTracer` instead of creating a new one each time.


chrisvest (approved):

buffer-api/src/main/java/io/netty/buffer/api/internal/LifecycleTracer.java
```diff
@@ -45,6 +45,8 @@ public static LifecycleTracer get() {
     public abstract <E extends Throwable> E attachTrace(E throwable);
 
     private static final class NoOpTracer extends LifecycleTracer {
+        private static final NoOpTracer INSTANCE = new NoOpTracer();
```
normanmaurer, line 48: nit: I would have made this package-private

PR #75, opened 2021-05-28
Make it possible for 3rd party PRs to get test results uploaded

chrisvest:


chrisvest:
I guess this won't take effect until it's merged.

